PR #59, opened 2026-04-15
Refactor `respond_challenge` to use asynchronous DNS setup

C0DEbrained:
The ACME proxy currently does a `sleep(30)` when the client requests the challenge. 
This means that some clients, such as Traefik, timeout when waiting for a response from UCM. 

This PR moves the DNS setup into a separate thread, so that the client will begin to poll immediately instead.

### DIFF
--- a/backend/services/acme/acme_proxy_service.py
+++ b/backend/services/acme/acme_proxy_service.py
@@ -253,7 +253,7 @@ class AcmeProxyService:
             if not provider:
                 raise Exception(f"No DNS provider configured for domain: {domain}. Configure it in ACME > Domains.")
             domain_providers[domain] = provider
-        The
+        
         # Forward to upstream Let's Encrypt
         payload = {
             "identifiers": identifiers,
@@ -392,7 +392,7 @@ class AcmeProxyService:
         # Find the proxy order that contains this challenge's authz URL
         order = self._find_order_for_challenge(chall_url, AcmeClientOrder)
         if not order:
-             raise RuntimeError("No pending proxy order found for this challenge.")
+            raise RuntimeError("No pending proxy order found for this challenge.")
 
         domain = order.domains_list[0].lstrip('*.')
         
@@ -482,7 +482,7 @@ class AcmeProxyService:
                 
                 # Trigger upstream validation
                 logger.info(f"[ACME Proxy BG] Triggering upstream validation for {domain}")
-                payload = {"keyAuthorization": key_authz}
+                payload = {}
                 resp = self._post_jws(chall_url, payload, kid=self.account_url)
                 
                 if resp.status_code != 200:
@@ -492,8 +492,9 @@ class AcmeProxyService:
                     
             except Exception as e:
                 logger.error(f"[ACME Proxy BG] Error in background challenge setup: {e}", exc_info=True)
-                if 'db' in locals():
-                    db.session.rollback()
+                db.session.rollback()
+            finally:
+                db.session.remove()
     
     def _find_order_for_challenge(self, chall_url, AcmeClientOrder):
         """Find the proxy order associated with a challenge URL."""
@@ -593,6 +594,9 @@ class AcmeProxyService:
             cert_url = order['certificate']
             cert_id = base64.urlsafe_b64encode(cert_url.encode()).rstrip(b'=').decode()
             order['certificate'] = f"{self.base_url}/acme/proxy/cert/{cert_id}"
+        
+        # Rewrite authorization URLs
+        
             
         return order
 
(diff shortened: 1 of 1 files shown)


C0DEbrained:
This code has been tested against a local instance of Traefik with the following configuration, and observed to work successfully:

```yaml
certificatesResolvers:
  local:
    acme:
      email: acme@my.email
      storage: /certs/acme-local.json
      caServer: "https://ucm.my.lan/acme/directory"
      httpChallenge:
        entryPoint: web
  external:
    acme:
      email: acme@my.email
      storage: /certs/acme.json
      caServer: "https://ucm.my.lan/acme/proxy/directory"
      dnsChallenge:
        provider: exec
        disablePropagationCheck: true
```

Note that the `exec` provider is configured via the environment variable `EXEC_PATH=/bin/true`, so that it is treated as a no-op.

NeySlim:
Thanks, will review and come back to you.

NeySlim (changes requested):
I found a few issues going through the diff:

**Blockers:**

1. **Line 256** : stray `The` word introduced (typo/merge artifact), will break syntax
2. **`challenge_data` and `resp` are undefined** : you moved the upstream `_post_jws()` call into the background thread, but the return block still references `challenge_data` and `resp.headers.get('Link')` which no longer exist in that scope. This will crash at runtime.

**Bugs:**

3. **`provider.get_zone_for_domain()` and `provider.get_acme_challenge_name()`** : these methods don't exist in our dns_providers. The original code uses `f"_acme-challenge.{domain}"` and passes `domain` directly. You'll need to use those or add the methods first.
4. **Challenge payload** : you changed `self._post_jws(chall_url, {}, ...)` to `{"keyAuthorization": key_authz}`. RFC 8555 §7.5.1 says the POST body for challenge response should be an empty JSON object `{}`. The keyAuthorization is computed by the server, not sent by the client.
5. **`if 'db' in locals()` in the except block** : `db` is a module-level import, it's always available. This check does nothing. Just call `db.session.rollback()` directly.
6. **`db.session.commit()` in background thread** : Flask-SQLAlchemy sessions are not thread-safe by default. You'll need a scoped session or create a new session for the bg thread. The `app.app_context()` alone doesn't guarantee a separate session.

**Minor:**

7. Line 395 has 5-space indent instead of 4

The approach is good, but the undefined variables in the return path (point 2) means it can't work as-is. Happy to help if you want to iterate on this.

C0DEbrained:
Thanks for the review, I'll admit I threw my agent at it and must have just hit a happy path where it all works.

Agent generated response: ;)
I have addressed the blockers and bugs identified in the PR review for acme_proxy_service.py. Below is a summary of the fixes applied:
Blockers & Syntax
- Fixed Syntax Error: Removed the stray word The at line 256 which was a merge artifact.
- Scope Correction: Verified that challenge_data and resp are correctly accessed within the appropriate scope in respond_challenge before spawning the background thread.

RFC Compliance & Logic

- RFC 8555 §7.5.1 Compliance: Updated the challenge response payload to be an empty JSON object {}. Previously, it incorrectly included keyAuthorization, which should be computed by the server.
- DNS Provider Methods: Confirmed that get_zone_for_domain() and get_acme_challenge_name() are present in BaseDnsProvider and available for use by the background thread.

Thread Safety & Database

- Thread Safety: Implemented db.session.remove() in a finally block within the background thread (_bg_respond_challenge) to ensure scoped sessions are properly cleaned up.
- Refined Error Handling: Removed the redundant if 'db' in locals() check in the exception block, as db is a module-level import.

Formatting

- Indentation: Fixed the 5-space indentation at line 395 to the standard 4 spaces.

All changes have been linted to ensure no new syntax or obvious logical errors were introduced.

NeySlim:
> Thanks for the review, I'll admit I threw my agent at it and must have just hit a happy path where it all works.

Don't we all do the same ? :D 

Reviewing now

NeySlim (commented):
Checked the new commit. Most points are fixed, good.

I need to correct myself on two things from my first review:

- `challenge_data` and `resp`: they ARE defined earlier in the function from the initial challenge fetch (lines ~371-377). I was looking only at the diff context and missed that. My bad.
- `get_zone_for_domain()` / `get_acme_challenge_name()`: they do exist in `BaseDnsProvider`, you were right.

What's fixed:
- Stray `The` removed
- Payload back to `{}` (RFC compliance)
- `db.session.remove()` in finally block
- Removed useless `if 'db' in locals()`
- Indentation

One thing left:

`get_order()` line ~596: there's an empty `# Rewrite authorization URLs` comment with nothing after it. Either implement the rewriting or remove the comment. Right now authorization URLs in the order response still point to upstream, which could break clients going through the proxy.

C0DEbrained:
Whoops, that stray comment is from my next PR I'll be opening 😁

NeySlim:
just remove that stray comment from this PR before merge so it stays clean. The authorization URL rewriting can come in your next one.

NeySlim (approved):
All good, thanks for the cleanup.

C0DEbrained:
Thanks!

PR #62, opened 2026-04-15
Proxy Authorization URLs in get_order and finalize_order endpoints

C0DEbrained:
Currently, the `get_order` and `finalize_order` endpoints do not proxy the authorization URLs. Therefore, some stateless acme clients will attempt to initiate the authorize challenge directly with let's encrypt. 

This PR modifies the response of these endpoints to replace the URLs with their proxied alternativesl


NeySlim:
I changed base to dev, so I don't have to double check merge.
Approved.

PR #87, opened 2026-04-24
fix: append UTC Z suffix to all datetime isoformat() serialisation

Hemsby:
All datetime fields stored in the database are naive UTC datetimes. Python isoformat() serialises them without a timezone indicator (e.g. "2027-04-24T08:36:27"), causing JavaScript to parse them as local time instead of UTC. This makes the configured timezone setting have no visible effect, meaning displayed times appear identical to UTC.

Append Z to all naive datetime isoformat() calls in to_dict() methods and API response constructors across the backend, making the UTC intent explicit in the serialised output and allowing the frontend Intl.DateTimeFormat formatter to correctly convert to the users configured timezone.

Affected files: models/__init__.py, api/v2/dashboard.py, api/v2/account.py, api/v2/acme.py, api/v2/cas.py,
api/v2/mtls.py, api/v2/settings.py, api/v2/sso.py, api/v2/truststore.py, api/v2/user_certificates.py


NeySlim (changes requested):
## Review

Premise is correct: SQLite stores naive UTC (`utc_now()` strips tzinfo), `isoformat()` then loses the UTC marker, and JS `new Date(...)` parses naive ISO strings as **local time** in some engines — breaking the timezone setting in the UI. Appending `Z` is a valid fix for naive-UTC datetimes.

There is **one real bug** and a couple of follow-up concerns.

---

### 🔴 HIGH — invalid ISO 8601 produced (`...+00:00Z`)

**`backend/api/v2/sso.py:990`**
```python
'not_after': cert.not_valid_after_utc.isoformat() + 'Z' if hasattr(cert, 'not_valid_after_utc') else cert.not_valid_after.isoformat(),
```

`cert.not_valid_after_utc` comes from the `cryptography` library and is **already timezone-aware** (UTC). Its `isoformat()` returns `2026-04-24T10:16:32+00:00`. Concatenating `'Z'` produces:

```
2026-04-24T10:16:32+00:00Z   ← not a valid RFC 3339 / ISO 8601 string
```

This is the only line in the diff where the source is an *aware* datetime. Everywhere else (`User`, `Session`, `Certificate`, `CA`, `AcmeAccount`, `TrustCert.not_after`, `NotificationLog`, …) the field comes from a SQLAlchemy column populated via `utc_now`, which is naive — those are fine.

**Fix:** drop the `+ 'Z'` for that one branch (the `+00:00` already encodes UTC), or normalise both branches:

```python
def _iso_utc(dt):
    if dt is None: return None
    if dt.tzinfo is not None:
        return dt.astimezone(timezone.utc).isoformat().replace('+00:00', 'Z')
    return dt.isoformat() + 'Z'
```

---

### 🟡 MEDIUM — fragile pattern, will silently regress

`isoformat() + 'Z'` is repeated 51× in this PR. Every future `to_dict()` or response builder will need to remember the suffix, and the next reviewer can't tell from the diff alone whether the source is naive (needs `Z`) or aware (must NOT have `Z`).

Strongly recommend introducing a single helper in `backend/utils/datetime_utils.py`:

```python
def utc_isoformat(dt):
    """ISO 8601 UTC string ('...Z') from naive-UTC or aware datetime, or None."""
    if dt is None:
        return None
    if dt.tzinfo is None:
        return dt.isoformat() + 'Z'
    return dt.astimezone(timezone.utc).isoformat().replace('+00:00', 'Z')
```

…and use it everywhere instead of inline concatenation. That removes the class of bug above and makes future audits trivial (`grep utc_isoformat` vs. `grep "isoformat()"`).

---

### 🟡 MEDIUM — incomplete coverage

The diff fixes ~51 occurrences but **~140 `isoformat()` calls without `Z` remain** in `api/v2/` and `models/` — the fix-or-don't decision will be lost as soon as this PR merges. Notable still-broken user-facing fields:

- `api/v2/certificates.py:723` (`valid_to` written into a record), `:1639` (audit details)
- `api/v2/policies.py:199-200` (`valid_from` / `valid_to` of generated policy)
- `api/v2/system.py:1521` (`started_at` system info)
- `models/crl.py`, `models/ocsp.py`, `models/policy.py`, `models/api_key.py`, `models/sso.py`, `models/email_notification.py`, `models/discovered_certificate.py`, `models/ssh.py`, `models/msca.py`, `models/webauthn.py`, `models/rbac.py`, `models/hsm.py`, `models/auth_certificate.py`, `models/truststore.py` — most `to_dict()` methods on these models still serialize without `Z`.

If the helper above is introduced, sweeping the rest becomes a mechanical follow-up.

For session-state fields (`auth.py`, `auth_methods.py`, `sso.py` `session['last_activity']` etc.) the missing `Z` is harmless because they're never re-parsed by JS — those can be left out of scope explicitly.

---

### Validation

Did not re-run the full test suite for this review, but verified the bug locally:

```
>>> aware = datetime.now(timezone.utc)
>>> aware.isoformat() + 'Z'
'2026-04-24T10:16:32.553544+00:00Z'   # invalid
>>> naive = aware.replace(tzinfo=None)
>>> naive.isoformat() + 'Z'
'2026-04-24T10:16:32.553544Z'          # correct
```

---

### Decision

**Request changes** — please:
1. Fix `sso.py:990` (drop the spurious `+ 'Z'` on the aware-datetime branch, or use the helper).
2. (Strongly preferred) introduce `utc_isoformat()` and migrate the diff to it; we can sweep the remaining 140 occurrences in a follow-up.

Once #1 lands the rest is non-blocking and can ship as-is.

Hemsby:
Updated — introduced utc_isoformat() helper and migrated all calls to use it. Also fixed the sso.py aware-datetime bug you flagged.

NeySlim (changes requested):
Excellent follow-up — the helper landed exactly as suggested and all 51 inline `+ 'Z'` sites were migrated cleanly. Tests pass (1558 ✅). The original HIGH at `sso.py:990` is correctly handled now (TZ-aware `cert.not_valid_after_utc` goes through `utc_isoformat` which strips `+00:00` before appending `Z`).

A few small follow-ups before merge — none are blockers individually, but #1 is the same bug class the PR set out to fix, so worth catching now:

### HIGH — missed instance of the original bug

**`backend/api/v2/sso.py:760`**
```python
'cert_expires': cert.not_valid_after_utc.isoformat()
```
Same situation as line 990: `cert.not_valid_after_utc` is a TZ-aware datetime (cryptography lib), returned to the frontend without a `Z` marker. The browser will parse it as local time. Should be:
```python
'cert_expires': utc_isoformat(cert.not_valid_after_utc)
```

### LOW — consistency nit

**`backend/api/v2/sso.py:990`** — the `else` fallback for old cryptography (`<42`) still uses raw `.isoformat()`:
```python
utc_isoformat(cert.not_valid_after_utc) if hasattr(cert, 'not_valid_after_utc') else cert.not_valid_after.isoformat()
```
The legacy attribute returns a *naive* UTC datetime, so wrapping it in `utc_isoformat()` would also be safe and keeps the contract uniform:
```python
utc_isoformat(cert.not_valid_after_utc if hasattr(cert, 'not_valid_after_utc') else cert.not_valid_after)
```

### LOW — unrelated change in this PR

**`backend/api/v2/system.py:37`** — `size_mb = round(..., 2)` → `round(..., 1)` is unrelated to the UTC fix. Either drop it from this PR or call it out in the commit message so it's traceable.

### Out of scope (not blocking)

There are still ~140 `.isoformat()` calls in other models (`models/crl.py`, `ocsp.py`, `policy.py`, `api_key.py`, `discovered_certificate.py`, `ssh.py`, `msca.py`, `webauthn.py`, `rbac.py`, `hsm.py`, `auth_certificate.py`, `truststore.py`, `email_notification.py`, plus a few `api/v2/*.py`) that emit naive UTC without `Z`. With the helper in place, the follow-up sweep is mechanical. Happy to do it as a separate PR.

Once #1 is fixed I'm happy to approve.

NeySlim (approved):
Both fixes landed cleanly:

- `sso.py:760` now goes through `utc_isoformat()` (HIGH resolved)
- `sso.py:990` else-branch consolidated into a single `utc_isoformat()` call (LOW nit resolved, both branches now uniform)
- The `size_mb` change is justified in its own commit message — fine to keep

1558 backend tests pass. LGTM 🚀

Follow-up sweep for the ~140 remaining `.isoformat()` calls in other models can be done as a separate PR — purely mechanical now that the helper exists.

PR #88, opened 2026-04-24
fix(settings): correct Last Optimized datetime display

Hemsby:
Sorry — this field was missed in #87. Two issues fixed:

  - **Backend (`system.py`):** `optimize_db()` and `check_integrity()` were storing timestamps using `strftime('%Y-%m-%d
   %H:%M:%S UTC')` — a non-standard format that JavaScript's `new Date()` cannot reliably parse. Changed both to
  `utc_isoformat(utc_now())` to produce proper ISO 8601 strings (e.g. `2026-04-24T16:33:52.123456Z`).

  - **Frontend (`SettingsPage.jsx`):** `formatDate()` was being called at data load time, before the auth verify had
  resolved and updated the timezone store. This caused the timestamp to always render in UTC regardless of the
  configured timezone. Moved the call to render time, matching the pattern used for certificates in #87.


NeySlim (approved):
LGTM — clean follow-up to #87, same UTC pattern. Tests pass (1558/1558). Thanks for catching this!

PR #97, opened 2026-04-29
fix: graceful fallback when pg user lacks session_replication_role privilege

### LINKED ISSUE #96: psycopg2 error on switching database
UCM version: 2.140

### Description
1 user (admin) , 1 HSM (openbao) and 1 CA (not using the hsm) configured.
Database url tests ok (postgresql://ucm:xxxxxxxx@postgresql.xx.xx.xx:5432/ucm

<hr>

Was trying to migrate & switch database from SQLite to Postgresql using a newly created database created with: 

`CREATE DATABASE ucm OWNER ucm;`
(user ucm was created earlier)

This action results in the following on screen error: 
<br>
Image
<br>

<hr 

Another issue is when choosing to switch *without* migration (which *is* successful on the same empty DB) you lose access to the system probably because the admin account/password is not migrated so you get locked out.  Makes sense, kind of, but is just a bit inconvenient :-)

(Workaround: set a new admin password immediately after switching DB while you still have a active session)

Cheers,
Rem.

Hemsby:
## Problem

When migrating from SQLite to PostgreSQL, the migration fails with:

`
psycopg2.errors.InsufficientPrivilege: permission denied to set parameter "session_replication_role"
`

This is raised by `SET session_replication_role = 'replica'`, which requires **superuser or replication privileges** in PostgreSQL. Standard application users (e.g. a dedicated `ucm` user created with `CREATE USER ucm`...) do not have these privileges, making SQLite → PostgreSQL migration completely unusable for the typical setup.

Fixes #96.

## Root cause

The migration uses `SET session_replication_role = 'replica'` to bypass FK trigger checks during bulk data load. This is a superuser-only operation in PostgreSQL.

## Fix

Wrapped the `SET` call in a try/except. If it fails due to insufficient privileges, a warning is logged and migration continues without the FK bypass. This is safe because the data being copied from a working SQLite database is already internally FK-consistent — no FK violations are expected during the load.

If the user's PostgreSQL role does have superuser/replication privileges, behaviour is unchanged.

## Testing

- Reproduced the error with a non-privileged `ucm` PostgreSQL user
- After fix, migration completes successfully without the privilege error
- Verified migration with a superuser account still takes the fast path (FK bypass enabled)


NeySlim:
Thanks @Hemsby for raising this and for the partial fixes in #97 and #98 — both have been folded into a single consolidated change on `dev` (commit c049824d92a5ba52f6cd89984d922f32aef18285, closes #96).

The consolidated fix:
- **Bootstrap auth tables** (users, sessions, webauthn_credentials, pro_sso_*, pro_custom_roles, pro_role_permissions, groups, group_members, api_keys) to the new target *before* the URL is persisted, so the next boot has a usable admin (extends #98 which only copied `users`).
- **Topological FK ordering fallback** when `SET session_replication_role` requires PG SUPERUSER (UCM may run with a least-privileged role) — addresses the FK-order point in #97 more robustly.
- **JSON column handling** for SQLite → PostgreSQL (db.JSON stored as TEXT in SQLite needs dict/list for psycopg2's JSON adapter).
- **Force-register all model packages** so `create_all()` sees every table.
- **NULL-safe `pg_get_serial_sequence`** in `_reset_pg_sequences`.
- **Audit log moved before service restart trigger** so the switch is recorded even if the restart kills the request.
- **14 new unit tests** in `backend/tests/test_database_admin.py`.

Closing this PR in favor of the consolidated commit. Your contribution made the regression visible — much appreciated 🙏

PR #98, opened 2026-04-29
fix: preserve user accounts when switching backend without migration

### LINKED ISSUE #96: psycopg2 error on switching database

Hemsby:
## Problem

When using "Switch (no migration)" to move to a new empty database, the application restarts pointing at a database with no users. The admin is immediately locked out with no straightforward recovery path.

Reported in #96 (second issue). The workaround — resetting the admin password before the service restarts — is fragile and non-obvious.

Closes #96.

## Root cause

The switch endpoint only persisted the new DATABASE_URL and triggered a restart. It made no attempt to prepare the target database, so the app started on a completely empty schema with no accounts.

## Fix

Before persisting the URL and restarting, a new `bootstrap_users_to_target()` function:

1. Creates the full schema on the target via SQLAlchemy `create_all()`
2. Creates the `_migrations` tracking table and marks all current migrations as applied (so the startup migration runner treats it as up-to-date, not a legacy install)
3. Copies the `users` table from the current database to the target
4. Resets PostgreSQL sequences so future inserts don't collide with copied IDs

The result: after the switch, the user can log in with their existing credentials. No certificates, CAs, or other data are carried over — this is still a "clean slate" switch for everything except authentication.

The confirm dialog message is also updated to clearly communicate this behaviour.


NeySlim:
Thanks @Hemsby for raising this and for the partial fixes in #97 and #98 — both have been folded into a single consolidated change on `dev` (commit c049824d92a5ba52f6cd89984d922f32aef18285, closes #96).

The consolidated fix:
- **Bootstrap auth tables** (users, sessions, webauthn_credentials, pro_sso_*, pro_custom_roles, pro_role_permissions, groups, group_members, api_keys) to the new target *before* the URL is persisted, so the next boot has a usable admin (extends #98 which only copied `users`).
- **Topological FK ordering fallback** when `SET session_replication_role` requires PG SUPERUSER (UCM may run with a least-privileged role) — addresses the FK-order point in #97 more robustly.
- **JSON column handling** for SQLite → PostgreSQL (db.JSON stored as TEXT in SQLite needs dict/list for psycopg2's JSON adapter).
- **Force-register all model packages** so `create_all()` sees every table.
- **NULL-safe `pg_get_serial_sequence`** in `_reset_pg_sequences`.
- **Audit log moved before service restart trigger** so the switch is recorded even if the restart kills the request.
- **14 new unit tests** in `backend/tests/test_database_admin.py`.

Closing this PR in favor of the consolidated commit. Your contribution made the regression visible — much appreciated 🙏

PR #99, opened 2026-04-30
fix(dashboard): chart cards overflow grid boundary + System Health missing scroll

Hemsby:
## Problems

Two layout bugs in the dashboard grid:

### 1. Certificate Activity & Status Distribution charts overflow card boundary

When resizing to a smaller viewport, the chart SVG breaks out of its card and renders behind other widgets. The card body had `min-h-[180px]` which, combined with the ~44px card header, forced the card taller than the grid slot at small sizes. The chart wrapper used a fixed `height:200` that compounded the problem — the chart rendered at full height regardless of available space.

### 2. System Health card silently clips content

The System Health card body used `overflow-hidden`, so when the card was resized small the service status badges were cut off with no scroll. Every other list-based card on the dashboard uses `overflow-y-auto`.

## Fix

**Charts:** Replaced `min-h-[180px]` with `overflow-hidden` on the card body (removes the minimum height floor that caused the overflow) and changed the chart wrapper from `height:200` to `height:'100%'` so the chart scales to fit available space rather than insisting on a fixed pixel height.

**System Health:** Changed `overflow-hidden` → `overflow-y-auto` so the card scrolls when content doesn't fit, consistent with all other dashboard cards.


Hemsby:
No rush for this, it was just a visual annoyance and I have already patched my own instance 😂

NeySlim (approved):
Clean fix. Two real layout bugs:

1. `min-h-[180px]` combined with fixed `height: 200` on the chart wrapper forced the card taller than its grid slot at small viewports — SVG escaped the card. Replacing with `overflow-hidden` + `height: 100%` is the right call: the chart now scales to the available flex space.
2. `overflow-hidden` on System Health was inconsistent with every other list-card on the dashboard, which use `overflow-y-auto`. Fix aligns it with the rest.

Parent Card has `h-full flex flex-col`, so `flex-1` + `height: 100%` resolves correctly. Will merge and ship in 2.142.

Thanks!

NeySlim:
I also wanna thank you personally for taking time testing and fixing things, it helps me a lot.
I think I'll pause new features for now, there are already plenty of them and only concentrate on fixes.
I went fast forward and didn't properly test it at maximum, the admin lock out after bad user postgres migration is a perfect example.

Hemsby:
I really like the product and so, im happy to invest the time and effort to help.  Found another issue so have raised a PR.

Kind Regards,

Roy Hagland

mail: ***@***.******@***.***>

Please note that electronic mail may be monitored. This e-mail is confidential. If you received it by mistake, please let me know and then delete it from your system; you should not copy, disclose, or distribute its contents to anyone nor act in reliance on this e-mail, as this is prohibited and may be unlawful.

Issue #147, opened 2026-06-30
ACME client always issues from Let's Encrypt, ignoring configured custom CA directory (e.g. Actalis) + EAB


fredlubrano:
## Summary

When a **custom ACME CA directory** is configured in *Settings → ACME client* (`acme.client.directory_url`, e.g. Actalis `https://acme-api.actalis.com/acme/directory`) together with EAB credentials, certificate issuance still goes to **Let's Encrypt** instead of the configured CA. The custom directory and the EAB stored in `SystemConfig` are never used by the issuance/renewal path.

**Version:** 2.179 (DEB install, SQLite)

## Steps to reproduce

1. In *Settings → ACME client*, set a custom `directory_url` to an EAB-required CA (e.g. Actalis), fill `eab_kid` / `eab_hmac_key`, set environment = `production`.
2. Register / verify the account (EAB looks fine in the UI).
3. Request a certificate (DNS-01).

## Expected

The order is created against the configured custom directory (Actalis), using the EAB credentials.

## Actual

The order is created against `https://acme-v02.api.letsencrypt.org/directory` and the certificate is issued by Let's Encrypt. No `AcmeClientAccount` row is ever created for the custom CA.

Observed in DB:
- `acme_client_accounts` only contains the Let's Encrypt Production row (`is_default=1`).
- `system_config` contains `acme.client.directory_url = <custom CA>`, `acme.client.eab_kid`, `acme.client.eab_hmac_key`.
- `acme_client_orders`: the order's `order_url` points to `letsencrypt.org`.

## Root cause

The request/renewal flow only ever passes an `environment` (`staging`/`production`) to `AcmeClientService`, and `_resolve_account()` maps both of those **unconditionally to Let's Encrypt URLs**:

- `backend/api/v2/acme_client/orders.py` — `request_certificate()` resolves `environment` from request/`acme.client.environment` and calls `AcmeClientService(environment=environment)`.
- `backend/services/acme/acme_client_service.py` — `_resolve_account()` maps `production` → `LE_PRODUCTION_URL`, `staging` → `LE_STAGING_URL`. The only branch that would use a custom default account is the one where **both** `environment` and `directory_url` are `None`, which never happens from the request API.
- The frontend `RequestCertificateForm.jsx` only exposes a binary `staging`/`production` selector — there is no way to pick a custom CA for issuance.

[...]

PR #148, opened 2026-06-30
fix(acme): route issuance to configured CA (Actalis) + complete CA chain & naming on export

fredlubrano:
## Summary

Two related fixes discovered while configuring **Actalis** as a custom ACME CA (with EAB) on a production install:

- **ACME issuance ignored the configured custom directory.** Issuance/renewal always resolved a Let's Encrypt environment, so certificates were issued by Let's Encrypt even though `acme.client.directory_url` pointed to Actalis. EAB credentials were also stored only in `SystemConfig` and never applied to the account row, so EAB-gated CAs could not authenticate.
- **Chain-aware exports dropped the CA chain for ACME/imported certs** and gave CA bags no friendly name.

### `acme-client`
- New `AcmeClientService.for_issuance()`: prefers a non-LE custom `directory_url` from `SystemConfig`, falls back to the standard environment otherwise.
- Backfills EAB (`acme.client.eab_kid` / `eab_hmac_key`) from legacy `SystemConfig` onto the `AcmeClientAccount` row when missing.
- Routes order creation/polling, account registration and renewal through `for_issuance()`.
- Adds `TestForIssuanceCustomDirectory` regression coverage.

### `cert-export`
- `_build_ca_chain()`: prefer the inline stored chain (ACME full-chain in `crt`), then walk managed CAs via `caref`, then complete the chain from locally known issuers (managed CAs + Trust Store) up to a self-signed root.
- `_named_cas()`: wrap CA certs as `PKCS12Certificate` with `friendlyName = CN` (fallback subject DN), so keystores show real CA names instead of `<leaf>_ca_N`.
- Uses the full chain + named CAs in the PKCS12, PFX and JKS paths.

## Test plan
- [x] `pytest backend/tests/test_acme_client_account_refactor.py` → 12 passed, 1 skipped
- [x] Verified on a live install: certificate now issued by Actalis (was Let's Encrypt).
- [x] Exported P12 reloaded via `cryptography.load_pkcs12`: full Actalis chain present, friendlyNames = CN of each CA (`Actalis Domain Validated TLS Server ECC CA 2025`, `Actalis TLS Server ECC Root CA 2025`, `Actalis Authentication Root CA`).

> Note: XCA's import dialog still labels CA bags `<leaf>_ca_N` — that is XCA-side behaviour (it ignores per-CA friendlyName); `keytool`/`certmgr`/`openssl` show the CN aliases correctly.

### DIFF
--- a/backend/api/v2/acme_client/account.py
+++ b/backend/api/v2/acme_client/account.py
@@ -47,7 +47,7 @@ def register_account():
         return error_response('Environment must be staging or production', 400)
 
     try:
-        client = AcmeClientService(environment=environment)
+        client = AcmeClientService.for_issuance(environment=environment)
         success, message, account_url = client.register_account(email)
 
         if success:
--- a/backend/api/v2/acme_client/orders.py
+++ b/backend/api/v2/acme_client/orders.py
@@ -203,7 +203,7 @@ def request_certificate():
 
     # Create order
     try:
-        client = AcmeClientService(environment=environment)
+        client = AcmeClientService.for_issuance(environment=environment)
         success, message, order = client.create_order(
             domains=domains,
             email=email,
@@ -301,7 +301,7 @@ def verify_challenges(order_id):
     force = bool(data.get('force'))
 
     try:
-        client = AcmeClientService(environment=order.environment)
+        client = AcmeClientService.for_issuance(environment=order.environment)
 
         results = {}
         challenges = order.challenges_dict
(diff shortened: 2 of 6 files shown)


NeySlim:
Thanks for the detailed report and follow-up! The ACME-routing half of this PR (for_issuance() + EAB backfill + routing all issuance entry points through it) shipped independently in v2.180 (#147 fix). The certificate-export half — using the inline stored chain, completing it from managed CAs + Trust Store, and naming CA bags with their CN — has been cherry-picked onto dev in commit `3129e9ac` (`feat(export): use full stored chain + friendly CA names in P12/PFX/JKS`), so this PR is now fully accounted for upstream. Closing as merged-in-pieces.

PR #155, opened 2026-07-03
fix(settings): Fix CORS logic for port 443

ellisdickinson46:
## Problem
When setting the `HTTPS_PORT` environment variable to the scheme standard (443). CORS Origins fail to match as browsers omit the port when sending the Origin header. This causes WebSocket connections to silently be rejected by the server as they fail the Origin check in Socket.IO.

There is currently a workaround to use the `CORS_EXTRA_ORIGINS` environment variable.

## Fix
Implement a check for the scheme standard port of 443 and only append the port number to the `_cors_origins` entries if they are non-standard.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Starting with v2.48, UCM uses Major.Build versioning (e.g., 2.48, 2.49). Earlier
 
 
 ## [Unreleased]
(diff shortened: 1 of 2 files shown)


NeySlim:
Thanks for the fix — valid catch (browsers omit the default port from the `Origin` header). Cherry-picked onto `dev` in commit d90739fa, will land in the next release. Closing in favor of the dev branch.

PR #172, opened 2026-07-06
fix(acme): DNS propagation timeout=0 and resolver logs (#171)

### LINKED ISSUE #171: Bug Report: ACME DNS-01 order stuck at "validating" because propagation check reports one resolver as perpetually "pending" (Quad9 9.9.9.9), even though that resolver answers correctly; dns_propagation_timeout=0 does not bypass the check
# Bug Report: ACME DNS-01 order stuck at "validating" because propagation check reports one resolver as perpetually "pending" (9.9.9.9), even though that resolver answers correctly; dns_propagation_timeout=0 does not bypass the check

## Summary

When issuing a certificate via the ACME client with a manual DNS provider and the DNS-01 challenge, the order never progresses to submission at the CA. The internal DNS propagation check repeatedly reports one of the three hardcoded public resolvers as pending (in this case 9.9.9.9), while the other two report OK. Because the check appears to gate on all three resolvers agreeing, the order stays in a pending/validating loop indefinitely and the challenge is never submitted to the ACME server for validation.

Critically, the resolver being reported as pending is reachable from the container and returns the correct TXT record when queried directly. This indicates the propagation checker is misclassifying a successful response as pending rather than a genuine DNS or network problem.

Setting the DNS propagation timeout to 0 (documented as "submit immediately") does not bypass the propagation check.

## Environment

UCM v2.184 (Docker image, :latest), migration 049 applied.
ACME proxy mode with upstream Let's Encrypt Production, EAB configured.
DNS-01 challenge with a manual DNS provider.
The order covers a wildcard plus its apex (two identifiers on one order).
The authoritative DNS for the domain is self-hosted (a primary plus a secondary), not the registrar.

## Observed behaviour

The ACME client logs show, on every poll cycle, that the authoritative resolver confirms the record, but the public propagation line consistently shows one resolver pending:

```
[utils.dns_txt_lookup] INFO DNS TXT confirmed for _acme-challenge.<domain> via authoritative resolver
[utils.dns_txt_lookup] INFO DNS public propagation for _acme-challenge.<domain>: 9.9.9.9=pending, 8.8.8.8=OK, 1.1.1.1=OK
[api.v2.acme_client.orders] INFO DNS TXT confirmed for <domain> (_acme-challenge.<domain>)
```

This repeats indefinitely. There is no subsequent log line indicating the challenge was submitted to the ACME server for validation, no authorization/challenge status transition, and no finalize/issue step. The order remains pending, and the UI shows the order stuck at "validating."

The 9.9.9.9=pending status persists across:
- multiple minutes and multiple poll cycles
- a full container restart (docker compose restart)
- setting dns_propagation_timeout to 0

## Proof that this is not a DNS or network problem

The TXT record is correct and fully propagated. Queried from outside the container, all three public resolvers return the expected value:

```
dig @8.8.8.8 _acme-challenge.<domain> TXT +short   -> correct value
dig @1.1.1.1 _acme-challenge.<domain> TXT +short   -> correct value
dig @9.9.9.9 _acme-challenge.<domain> TXT +short   -> correct value
```

More importantly, from inside the UCM container itself, a raw DNS query to 9.9.9.9:53 succeeds and returns the same answer as 8.8.8.8. Because the container lacks dig/nslookup, this was verified with a minimal Python UDP DNS query:

```
# docker exec <container> python3 <minimal DNS query script>
9.9.9.9 ANSWERED 96 bytes
8.8.8.8 ANSWERED 96 bytes
```

Both resolvers returned an identical 96-byte response from within the container. UDP and TCP reachability to 9.9.9.9:53 from the container were also confirmed. Therefore the container can reach 9.9.9.9, that resolver answers, and the answer matches the other resolvers. Despite this, UCM's propagation checker reports 9.9.9.9 as pending.

This strongly suggests the propagation checker is either not correctly parsing or comparing the 9.9.9.9 response, or is applying a per-resolver timeout/retry that fails for this resolver specifically, and is then treating that as pending rather than success.

## Impact

For a manual DNS provider, this makes it effectively impossible to complete issuance when any one of the three hardcoded resolvers is misclassified, because:

1. the check appears to require all three public resolvers to report OK, and
2. dns_propagation_timeout=0 does not bypass the check as its help text implies.

The order cannot be pushed to the CA even though the record is correct everywhere and the authoritative resolver is already confirmed by UCM itself. Note that Let's Encrypt validates against the domain's authoritative nameservers, not against the public recursive resolvers, so the public-resolver propagation check is a convenience pre-check rather than a correctness requirement; blocking issuance on it is stricter than the CA requires.

## Suggested fixes (any one would resolve the block)

1. Fix the evaluation of the 9.9.9.9 response so a correct answer is classified as OK rather than pending. Since a raw query to 9.9.9.9 from the container returns the correct record, the query itself works; the parsing/timeout handling for that resolver appears to be the issue.

2. Make dns_propagation_timeout=0 genuinely submit immediately, bypassing the public propagation pre-check as the help text states. This gives operators an explicit escape hatch.

3. Do not require all three public resolvers to agree. Proceeding when the authoritative resolver is confirmed (which UCM already logs as "DNS TXT confirmed via authoritative resolver"), or when a majority of resolvers report OK, would be sufficient and matches what the CA actually checks.

4. Allow the set of propagation-check resolvers to be configured (or individual resolvers to be disabled), so a single problematic resolver cannot block issuance.

## Workaround in the meantime

None found within UCM: timeout=0 does not bypass the check, and a container restart does not clear the pending state. Issuance via UCM's ACME client is blocked for manual DNS providers whenever this misclassification occurs.

fredlubrano:
## Summary

- **`dns_propagation_timeout=0`** now skips the DNS propagation gate for **manual Verify** as well as auto DNS (previously manual Verify still blocked on a single failed lookup).
- **Manual Verify with timeout > 0** polls up to the configured value (was a single pass with `timeout=0` internally).
- **Resolver logs** show per-public-resolver pending reasons (`nxdomain`, `timeout`, `value_mismatch`) and a summary clarifying that 3/3 is **not** required — authoritative or any resolver is enough.
- **TXT matching** accepts concatenated multi-string RDATA (Quad9-style split chunks).
- **Docs + i18n** aligned on auto vs manual behavior; `docs/testing/ACME-DNS-PROPAGATION.md` documents `acme.dns01_nameservers` to exclude a problematic resolver.

Fixes #171

## Test plan

- [x] `pytest backend/tests/test_acme_dns_selfcheck.py` (17 passed)
- [ ] Manual DNS order with `dns_propagation_timeout=0` → Verify submits without `dns_not_ready`
- [ ] Manual DNS with timeout=120 → Verify waits/polls before blocking
- [ ] Auto DNS with timeout=0 → background path skips polling and submits
- [ ] Logs show `9.9.9.9=pending (nxdomain)` style lines and `N/3 resolvers OK (informational only…)`


NeySlim:
Closing this PR in favor of the work already shipped on `dev`.

**Why not merging:** this PR branched from `dev` before commit `40c892d5`, which already shipped the core of the #171 fix (auto-poll `timeout=0` bypass, exception-type logging in `check_public_resolvers`, "diagnostic, does not block issuance" wording). The three files this PR touches (`orders.py`, `dns_txt_lookup.py`, `test_acme_dns_selfcheck.py`) collide with that commit — the PR is currently `mergeable_state: dirty` and can't be merged as-is.

**What was kept:** all three genuinely-new pieces from this PR have been integrated into `dev` in commit `bb4b2a20`, with attribution preserved in that commit's message:

1. `_txt_rdata_matches` now joins multi-string TXT RDATA before comparing (RFC 1035 §3.3.14; Quad9 splits long ACME tokens across quoted strings) — the likely actual root cause of the reporter's `9.9.9.9=pending`.
2. `timeout=0` bypass now also covers the **manual Verify** path (`verify_challenges`), not just the auto-poll background flow; manual Verify with `timeout > 0` polls up to the configured value instead of a single arbitrary pass.
3. `docs/testing/ACME-DNS-PROPAGATION.md` and the rewritten `dnsPropagationTimeoutHelper` i18n (9 locales).

**What was dropped as duplicate:** `_resolver_failure_reason` categorical labeling was not kept — `check_public_resolvers` on `dev` already logs the raw exception type at DEBUG (`NXDOMAIN` / `Timeout` / `ConnectionError` / …) per failing resolver, which is strictly more informative than the categorical label. If anyone feels strongly that the INFO-level `9.9.9.9=pending (nxdomain)` format is worth having too, it can be layered on top of the existing DEBUG logging in a follow-up.

Thanks for the investigation — the TXT chunk-concatenation catch in particular was the piece that turned "interesting diagnostic improvement" into "actual root-cause fix."

PR #175, opened 2026-07-09
feat(acme): LOT A — shared DNS self-check for proxy/renewal

fredlubrano:
## Summary

This PR completes **ACME LOT A**: proxy and renewal no longer rely on a hardcoded 30s sleep after publishing the DNS-01 TXT record. They now use the **same active DNS poll** as the ACME client auto-poll, driven by `acme.client.dns_propagation_timeout`.

**Verbose ACME/DNS logging** (`acme.client.debug_logging`) was split out to **#176** per review feedback (separate PR, no utils→services coupling).

### Problem

| Path | Before | Issue |
|------|--------|-------|
| ACME client auto-poll | Active DNS poll (`dns_propagation_timeout`) | OK |
| ACME proxy (`_bg_respond_challenge`) | `sleep(30)` fixed | Ignored operator timeout; not equivalent to client behavior |
| Auto-renewal | `sleep(30)` fixed | Same |
| Upstream submit on missing TXT | Always after sleep | Could burn Actalis/LE tokens when TXT not yet visible |

### Solution

1. **Shared module** `services/acme/dns_selfcheck.py`
   - `dns_propagation_timeout()` — reads `acme.client.dns_propagation_timeout` (default 120s, 0 = skip)
   - `wait_for_challenges()` / `wait_for_txt()` — poll every 5s until TXT visible or timeout
   - `challenge_txt_name()` — fallback when `dns_txt_name` is absent (renewal rows)

2. **Proxy** (`acme_proxy_service.py`)
   - Replaces `sleep(30)` with `wait_for_txt()`
   - If timeout elapses without TXT → **no upstream submit**, challenge marked `dns_not_ready`, DNS record cleaned up

3. **Renewal** (`acme_renewal_service.py`)
   - Same active poll; raises if TXT not ready
   - **`try/finally` cleanup** — deletes DNS TXT records on propagation or finalization failure (review fix)

4. **Client** (`orders.py`)
   - Refactored to delegate to shared `dns_selfcheck` (single implementation)

### Settings reference

| UI label | Config key | Default | Notes |
|----------|------------|---------|-------|
| Délai de propagation DNS (s) | `acme.client.dns_propagation_timeout` | 120 | Max poll time before submit (not a blind sleep) |
| Intervalle de poll (s) | `order_poll_interval_sec` (per CA account) | 3 | **Different** — CA order status poll *after* submit |

## Review feedback addressed (`7cf6f120`)

- Removed debug-logging toggle from this PR → **#176**
- Renewal cleans up DNS TXT on propagation failure (`try/finally`)
- `wait_for_challenges` uses `dns_txt_name` fallback (no `KeyError`)
- Added tests: `test_acme_renewal_dns_cleanup.py`, `challenge_txt_name` / missing `dns_txt_name` in `test_acme_dns_selfcheck.py`

## Documentation

- `CHANGELOG.md` — Unreleased entry
- `docs/testing/ACME-DNS-PROPAGATION.md` — LOT A proxy/renewal behavior, pytest plan

### Automated (CI / local)

```bash
cd backend
python -m pytest tests/test_acme_dns_selfcheck.py tests/test_acme_renewal_dns_cleanup.py -q
python -m pytest tests/test_acme_proxy_ca_account.py -q
python -m pytest tests/test_acme.py::TestAcmeClientSettings -q
```

### Manual — proxy happy path

- [ ] Set `Délai de propagation DNS (s)` to e.g. `200`
- [ ] Request a cert via ACME proxy (DNS-01, auto provider)
- [ ] Logs show: TXT create → poll started → TXT confirmed → upstream validation → cert stored → DNS cleanup
- [ ] Public resolver line: `9.9.9.9=OK, 8.8.8.8=OK, 1.1.1.1=OK` (diagnostic)

### Manual — proxy DNS not ready

- [ ] Simulate missing TXT (or mock `wait_for_txt` returning `ok: false`)
- [ ] Verify upstream challenge is **not** submitted
- [ ] Challenge status = `dns_not_ready`, DNS record cleaned up

### Manual — renewal

- [ ] Trigger auto-renewal with DNS-01
- [ ] Confirm active poll (not fixed 30s sleep) and cert re-issued
- [ ] On propagation failure, TXT records are deleted (no leak)

### Regression

- [ ] `dns_propagation_timeout=0` still skips wait on client auto-poll and manual verify (#171 behavior unchanged)
- [ ] Client auto-poll still honors full configured timeout in background thread

## Lab validation

Validated on a test UCM instance:
- `dns_propagation_timeout=200` persisted and read at runtime
- Proxy issuance for `test.ucm.pfcorp.eu` / `test1.ucm.pfcorp.eu`: authoritative + public resolvers OK, cert issued, DNS cleaned up

## Follow-up

- **#176** — verbose ACME/DNS logging toggle (`acme.client.debug_logging`)
- LOT B — HTTP-01 client / DNS-01 upstream (separate proposal)


NeySlim (changes requested):
Thanks for this one — the core of the PR is exactly the right direction. I ran the branch locally: full backend suite passes (2247 passed), and the new `dns_not_ready` proxy test covers the important new behavior. Replacing the blind 30s sleep in proxy/renewal with the same active poll as the client, honouring `dns_propagation_timeout`, and refusing to submit upstream when the TXT isn't visible is a real improvement (and saves upstream tokens on slow propagation). I want to merge the self-check part.

Three things need fixing first, and one design point I'd like to split out.

### Requested changes

**1. i18n sync (blocking, CI/pre-commit will reject)**
`acme.debugLogging` / `acme.debugLoggingDesc` are only added to `en.json` and `fr.json`. All 9 locale files must stay in sync — `node scripts/check-i18n-sync.js` currently fails on the branch with 2 missing keys in de/es/it/ja/pt/uk/zh. (This goes away by itself if the toggle moves to its own PR, see below.)

**2. TXT records leak on the new renewal failure path**
In `acme_renewal_service.py`, the new `raise Exception("DNS propagation not ready …")` fires *after* the TXT records were created, but DNS cleanup only happens on the success path (after finalize). Every renewal attempt that times out on propagation now leaves orphan TXT records at the DNS provider — and since the scheduler retries failed renewals, they accumulate. The proxy path does clean up on this failure (`_delete_dns_record`); the renewal path needs the same, e.g. a `try/finally` around the challenge section or cleanup in the failure branch.

**3. Potential `KeyError` in `wait_for_challenges`**
The shared poll accesses `c['dns_txt_name']` directly, but the renewal code three lines above uses `challenge_info.get('dns_txt_name', f"_acme-challenge.{…}")` — i.e. that key can legitimately be absent in renewal challenge dicts. If it is, the poll dies with `KeyError` instead of a clean not-ready result. Please apply the same fallback in `dns_selfcheck.wait_for_challenges()`.

### Design point — the debug logging toggle

I'd prefer to extract the `acme.client.debug_logging` toggle into a separate PR and discuss the pattern there, for two reasons:

- `acme_log()` runs a `SystemConfig` query **per log line, even with the toggle off**, inside the very loops it instruments (5s ticks × domains × 3 public resolvers, for up to `dns_propagation_timeout`). On a PostgreSQL backend that's a network round-trip per diagnostic line. It needs at least a short-TTL cache or a read-once-per-poll-session flag.
- It also makes `utils/dns_txt_lookup.py` import from `services/` (which pulls in the ORM) — low-level utils shouldn't depend on services. And it establishes a per-subsystem runtime log-level pattern that other modules will want to copy, so I'd rather settle that pattern deliberately (a runtime named-logger level control could do the same job with zero per-line cost).

The troubleshooting value is real — I just don't want to lock in the mechanism as a side effect of the DNS work.

### Timing

`v2.189-rc1` is already cut, so this will land in the next cycle — no rush on the rebase, and the "From v2.188+LOT-A" wording in `docs/testing/ACME-DNS-PROPAGATION.md` will need a version touch-up at merge time.

NeySlim (approved):
All three requested changes addressed, and thanks for extracting the debug toggle — happy to discuss that pattern in its own PR.

Verified locally on the updated branch:
- i18n sync check passes (toggle removal resolved it)
- full backend suite: 2244 passed
- renewal now cleans up TXT records on the propagation-failure path (`try/finally` + dedicated test)
- `challenge_txt_name()` fallback covers renewal dicts without `dns_txt_name`

One non-blocking observation for a possible follow-up: `dns_txt_created` is only set after *all* creates succeed, so if `create_txt_record` fails midway through a multi-domain order, the records already created before the failure still leak (pre-existing behavior, not a regression from this PR). Setting the flag once the first record is created would cover it, since the cleanup helper is already best-effort per record.

LGTM — approving.

fredlubrano:
Thanks for the review and approval.

Follow-up on the multi-domain observation: opened **#177** to set `dns_txt_created` after each successful `create_txt_record`, so partial create failures still trigger the existing `try/finally` cleanup.

Verbose debug logging remains in **#176** as discussed.

PR #177, opened 2026-07-09
fix(acme): renewal TXT cleanup on partial multi-domain create failure

fredlubrano:
## Summary

Follow-up to review note on **#175** (non-blocking): when a multi-domain renewal fails midway through `create_txt_record`, TXT records already created for earlier domains were not cleaned up because `dns_txt_created` was only set after **all** creates succeeded.

This PR sets `dns_txt_created = True` after each successful create so the existing `try/finally` cleanup runs on partial failures too.

## Test plan

- [x] `pytest tests/test_acme_renewal_dns_cleanup.py` (propagation failure + partial create failure)

## Dependency

Stacked on **#175** (`feat/acme-lot-a-dns-selfcheck`). Rebase onto `dev` after #175 merges for a clean diff.


fredlubrano:
Follow-up to the non-blocking note on **#175** — thanks for flagging it.

`dns_txt_created` is now set after each successful `create_txt_record` in the renewal loop, so the existing `try/finally` cleanup runs when a later domain fails mid-loop.

Added `test_renewal_deletes_txt_when_create_fails_on_second_domain` in `test_acme_renewal_dns_cleanup.py`.

NeySlim (approved):
Exactly the fix suggested in the #175 review — verified locally: full backend suite passes (2245) and the new multi-domain partial-failure test covers the leak. Thanks for the quick follow-up.

PR #178, opened 2026-07-10
Fix API-key creation rejecting valid permission resources (csrs, etc.)

heidrickla:
`POST /api/v2/account/apikeys` validates `permissions` against a hardcoded `valid_resources` list in `apikeys.py` that has drifted from `ROLE_PERMISSIONS` in `auth/permissions.py` — it's missing `csrs`, `user_certificates`, `templates`, `truststore`, `est`, `hsm`, `ssh`, `policies`, `approvals`, `key_recovery`, `audit`, `groups`, and lists two non-resources (`users`, `system`).

So an API key can't be scoped to those resources. For example `write:csrs` — required by `POST /api/v2/csrs/<id>/sign` — is rejected with `400 "Invalid permission resource: csrs"`, leaving CSR signing reachable only by a full `*` admin key.

**Fix:** derive the allowed resources from `ROLE_PERMISSIONS` (the source of truth `@require_auth` already uses) so the two can't drift again. The existing "can only grant permissions you already have" check is unchanged.

### DIFF
--- a/backend/api/v2/account/apikeys.py
+++ b/backend/api/v2/account/apikeys.py
@@ -5,7 +5,7 @@ from flask import request, g, current_app
 from models import db
 from models.api_key import APIKey
 from auth.unified import AuthManager, require_auth, has_permission
-from auth.permissions import get_role_permissions
+from auth.permissions import get_role_permissions, VALID_RESOURCES
 from services.audit_service import AuditService
 from utils.response import success_response, error_response, created_response
 from utils.db_transaction import safe_commit
@@ -63,7 +63,11 @@ def create_api_key():
     # Validate permissions format — users cannot grant permissions they don't have
     user_perms = get_role_permissions(g.current_user.role)
     valid_categories = ['read', 'write', 'delete', 'admin']
-    valid_resources = ['cas', 'certificates', 'acme', 'scep', 'crl', 'settings', 'users', 'system']
+    # Derived from ROLE_PERMISSIONS (single source of truth) so this validator can't drift
+    # from the resources @require_auth actually enforces — the previous hardcoded list was
+    # missing csrs/user_certificates/templates/truststore/est/hsm/ssh/policies/approvals/
+    # key_recovery/audit/groups and listed two non-resources (users/system).
+    valid_resources = VALID_RESOURCES
 
     for perm in data['permissions']:
         if perm == '*':
--- a/backend/auth/permissions.py
+++ b/backend/auth/permissions.py
@@ -53,16 +53,27 @@ ROLE_PERMISSIONS = {
 }
 
 
-# Canonical set of resources referenced anywhere in ROLE_PERMISSIONS. Derived from this
-# single source of truth so that permission validators (e.g. API-key creation) stay in
-# sync with the `action:resource` scopes actually enforced by @require_auth and can't
-# silently drift from them.
-VALID_RESOURCES = sorted({
-    perm.split(':', 1)[1]
-    for perms in ROLE_PERMISSIONS.values()
-    for perm in perms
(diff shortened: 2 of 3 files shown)


NeySlim (changes requested):
Thanks for the PR — the diagnosis is correct (`valid_resources` has drifted and `write:csrs` etc. can't be granted), and deriving the list from a single source of truth is the right idea. But `ROLE_PERMISSIONS` isn't the complete source of truth, so as-is this trades one drift bug for another.

**The issue:** `ROLE_PERMISSIONS` only enumerates permissions for non-admin roles — admin is just `['*']`, so admin-only resources never appear in it. The codebase's `@require_auth` decorators do use `admin:system` (~48 call sites), `read/write/admin/delete:users`, and `read/write/delete:sso`, and at enforcement time an API key's stored permissions are matched directly against those strings — so keys scoped to `users`/`system` are valid and functional today.

Deriving `VALID_RESOURCES` purely from `ROLE_PERMISSIONS` therefore:
1. **Regresses `users` and `system`**: an admin can no longer create a key scoped to e.g. `read:users` or `admin:system` (rejected with `400 Invalid permission resource`) — the same class of bug this PR fixes, in the other direction.
2. **Leaves `sso` unfixable**: it's used by `@require_auth` but absent from both the old hardcoded list and the derived set.

**Test failure:** running the backend suite on this branch, `tests/test_account.py::TestCreateAPIKey::test_cannot_grant_permissions_beyond_role` fails — it posts `['admin:system', 'write:users']` as a viewer and expects `403` (privilege-escalation check), but now gets `400` because `system` is rejected as an invalid resource before the escalation check runs.

**Suggested fix:** keep the derivation but union in the admin-only resources explicitly, e.g. in `permissions.py`:

```python
# Resources only reachable via the admin '*' wildcard — they never appear in
# ROLE_PERMISSIONS but are enforced by @require_auth decorators.
ADMIN_ONLY_RESOURCES = {'users', 'system', 'sso'}

VALID_RESOURCES = sorted({
    perm.split(':', 1)[1]
    for perms in ROLE_PERMISSIONS.values()
    for perm in perms
    if ':' in perm
} | ADMIN_ONLY_RESOURCES)
```

Please also add a regression test asserting that key creation accepts every resource in `VALID_RESOURCES` (and specifically `csrs`, `users`, `system`, `sso`) — that would have caught both the original drift and this one.

Two housekeeping points:
- Please retarget the PR to `dev` — `main` only receives release merges.
- The comment block in `apikeys.py` narrating what the previous list was missing can be trimmed; the commit message already covers the history.

NeySlim:
Thanks for the quick turnaround — the fix looks right now.

Verified on this branch: the full backend suite passes (2254 passed, 6 skipped), including the previously failing `test_cannot_grant_permissions_beyond_role`. The new AST-based regression test is a nice touch: it pins `VALID_RESOURCES` to the scopes `@require_auth` actually enforces, so this class of drift can't come back silently.

Two remaining points before merge:

1. **Please retarget the PR to `dev`** — `main` only receives release merges; this will ship with the next release cycle.
2. The comment in `apikeys.py` still says the old list "listed two non-resources (users/system)" — that turned out to be wrong (they're admin-only resources, as your `permissions.py` comment now correctly explains). Please drop or reword that part so the two comments don't contradict each other.

Happy to approve once those are done.

NeySlim (approved):
Approving — verified the full backend suite passes both on this branch and on a test merge with dev (2254 passed). The stale wording in the apikeys.py comment ("two non-resources") will be tidied up in a follow-up on dev. Thanks for the fix and the AST regression test!

heidrickla:
Np, I would rather follow the repo than spin off on my own.

PR #187, opened 2026-07-11
fix(security): GHSA auth/methods disclosure, ACME account SSRF, webhook test rebinding

fredlubrano:
## Summary

Private vulnerability reports (coordinated disclosure):

- [GHSA-p4hj-mmxv-xh65](https://github.com/NeySlim/ultimate-ca-manager/security/advisories/GHSA-p4hj-mmxv-xh65) — unauthenticated `mtls_user` leak via forged `X-SSL-Client-*` on `/api/v2/auth/methods`
- [GHSA-5p92-5vpr-2x5w](https://github.com/NeySlim/ultimate-ca-manager/security/advisories/GHSA-5p92-5vpr-2x5w) — ACME multi-CA account create/register SSRF to loopback/metadata
- [GHSA-q7j8-h9jm-qxw8](https://github.com/NeySlim/ultimate-ca-manager/security/advisories/GHSA-q7j8-h9jm-qxw8) — legacy settings webhook test DNS rebinding (`requests.post` without pin)

## Changes

| Area | Fix |
|------|-----|
| `auth_methods.py` | Same `is_request_from_trusted_proxy()` gate as `login_mtls()` before honouring proxy cert headers |
| `acme_client/accounts.py` | `validate_url_not_cloud_metadata()` on create + register |
| `acme_client_service.py` | `_fetch_directory()` via `safe_request_get()` (DNS-pinned) |
| `settings/webhooks.py` | `test_webhook()` via `safe_request_post()` (aligned with modern delivery) |

## Test plan

- [x] `pytest tests/test_auth_methods.py::TestDetectAuthMethods::test_auth_methods_ignores_spoofed_proxy_headers`
- [x] `pytest tests/test_acme_ca_timing.py::test_create_ca_account_rejects_loopback_directory`
- [x] `pytest tests/test_settings.py::TestWebhooks::test_test_webhook_uses_pinned_safe_request_post`

Please link this PR to the three GHSA drafts when accepting the reports.


NeySlim (changes requested):
Thanks for these — the three fixes are on point. One regression to fix before merge, plus a CHANGELOG entry.

### Regression — ACME directory fetch drops `verify_ssl`

In `services/acme/acme_client_service.py::_fetch_directory`, switching from `self.session.get(...)` to `safe_request_get(...)` loses the session's TLS setting. The session is built with `create_session(verify_ssl=self.verify_ssl)`, whereas `safe_request_get` starts a fresh `requests.get` defaulting to `verify=True`. An internal ACME CA with a self-signed cert and `acme.client.verify_ssl=false` (a supported, documented option) will now fail the directory fetch on TLS verification.

Please forward the setting (the `**kwargs` passes it through; keeping the User-Agent would be a bonus):

```python
resp = safe_request_get(
    self.directory_url,
    timeout=self._http_timeout(),
    verify=self.verify_ssl,
)
```

The SSRF pinning is still applied — this only restores the opt-out.

### CHANGELOG

Please add a `### Security` entry under `[Unreleased]` covering the three fixes (auth/methods proxy-header trust, webhook test DNS pinning, ACME account/directory SSRF).

The rest looks good: the trusted-proxy gate matches `login_mtls()`, `validate_url_not_cloud_metadata` is the right choice (it keeps RFC1918 internal CAs working while blocking loopback/metadata), and the added tests cover the regressions well.

fredlubrano:
Addressed in `adeebdee`:

- **`_fetch_directory()`** — forwards `verify=self.verify_ssl` and the session `User-Agent` to `safe_request_get()` so internal ACME CAs with `acme.client.verify_ssl=false` keep working; DNS pinning unchanged.
- **CHANGELOG** — `[Unreleased]` **Security** entry added for all three GHSA fixes (auth/methods trusted-proxy gate, legacy webhook test DNS pinning, ACME account/directory SSRF).

Regression tests still green on the three new cases. Ready for another look when you have a moment.

NeySlim (approved):
Both points addressed — thanks.

- `_fetch_directory` now forwards `verify=self.verify_ssl` (and preserves the User-Agent), so internal ACME CAs with `acme.client.verify_ssl=false` keep working while the DNS pinning stays applied. ✅
- CHANGELOG `### Security` entries added for all three fixes with GHSA references. ✅

Verified locally: `test_acme_ca_timing.py`, `test_auth_methods.py`, `test_settings.py` — 147 passing. LGTM.

PR #188, opened 2026-07-11
fix(security): GHSA import CA gate, SAML rebinding, search PII, auth enum

fredlubrano:
## Summary

Coordinated disclosure for four private vulnerability reports (triage):

- [GHSA-rgcp-9wxj-6896](https://github.com/NeySlim/ultimate-ca-manager/security/advisories/GHSA-rgcp-9wxj-6896) — **High** — smart import / OPNsense CA import bypasses `write:cas`
- [GHSA-9crx-4487-wvxc](https://github.com/NeySlim/ultimate-ca-manager/security/advisories/GHSA-9crx-4487-wvxc) — **Medium** — SAML IdP metadata fetch DNS rebinding
- [GHSA-4v84-cxgw-8g4g](https://github.com/NeySlim/ultimate-ca-manager/security/advisories/GHSA-4v84-cxgw-8g4g) — **Medium** — global search user PII without `read:users`
- [GHSA-xxhm-683g-fjj5](https://github.com/NeySlim/ultimate-ca-manager/security/advisories/GHSA-xxhm-683g-fjj5) — **Medium** — `/auth/methods` username enumeration via `mtls_certificates`

## Fixes

| GHSA | Change |
|------|--------|
| GHSA-rgcp-9wxj-6896 | Require `write:cas` when `import_cas` is true (smart import) or OPNsense import selects CA items |
| GHSA-9crx-4487-wvxc | Route SAML metadata fetch through `safe_request_get()` |
| GHSA-4v84-cxgw-8g4g | Gate search `users[]` on `read:users` |
| GHSA-xxhm-683g-fjj5 | Neutral pre-auth response (no per-user credential counts) |

## Test plan

- [x] `test_execute_ca_import_requires_write_cas`
- [x] `test_fetch_metadata_uses_pinned_safe_request_get`
- [x] `test_search_hides_users_without_read_users`
- [x] `test_post_methods_with_unknown_username` (enumeration oracle removed)

PR #190, opened 2026-07-11
fix(security): smart-import write:cas bypass for legacy CA certs without BasicConstraints

fredlubrano:
## Summary

Closes an incomplete fix from #188 (GHSA-rgcp): CA material with **keyCertSign** but **no BasicConstraints** was classified as a leaf certificate, allowing callers with only `write:certificates` to import CA private keys into the `certificates` table.

## Root cause

- Permission gate in `smart_import.py` used `analysis.summary['cas']`, which only counts `BasicConstraints CA:true`.
- `_import_certificates()` imported non-`is_ca` certs with matched private keys, including legacy roots that assert `keyCertSign` without BC.

## Fix

- Parse `keyCertSign` from KeyUsage in the smart-import parser.
- Treat `is_ca OR key_cert_sign` as CA import targets for summary counting, `_import_cas()`, and `_import_certificates()` exclusion.

### Automated (local — green on branch `fix/smart-import-ca-gate-keycertsign`)

```bash
cd backend
pytest tests/test_misc_routes.py::TestSmartImport -x -q
# 7/7 passed (2026-07-11)
```

| Test | Expected |
|------|----------|
| `test_execute_ca_import_requires_write_cas` | API key `write:certificates` only + content with CA → **403** |
| `test_execute_cert_only_import_allowed_without_write_cas` | API key cert-only + `summary.cas=0` → **not 403** (NeySlim #188 regression guard) |
| `test_execute_ca_import_passes_permission_with_write_cas` | API key with `write:cas` + CA content → gate passes |
| `test_execute_legacy_ca_without_basic_constraints_requires_write_cas` | Self-signed RSA, **keyCertSign + crlSign**, **no BasicConstraints**, cert+key bundle, `write:certificates` only → **403** |

Related gate alignment (unchanged, sanity check):

```bash
pytest tests/test_opnsense_import.py -k write_cas -q
# passed
```

### Manual API — attack scenario (403 expected before fix, blocked after)

1. Create API key with `read:certificates`, `write:certificates` — **no** `write:cas`.
2. Build or obtain a legacy-style PEM bundle: self-signed cert with KeyUsage `keyCertSign`/`cRLSign`, **without** BasicConstraints, plus matching private key.
3. `POST /api/v2/import/execute` with `options.import_cas=true` (default).
4. **Before fix:** import succeeds; CA private key stored under `certificates` (privilege escalation).
5. **After fix:** **403** `write:cas permission required to import CAs`.

### Manual API — legitimate cert-only import (must still work)

1. Same API key (no `write:cas`).
2. Import a normal server certificate + key (KeyUsage `digitalSignature` + `keyEncipherment`, no `keyCertSign`).
3. **Expected:** import succeeds; no `write:cas` required.

### Manual API — authorized CA import

1. API key with `write:certificates` + `write:cas`.
2. Import legacy CA bundle (keyCertSign, no BC) or standard CA (BC CA:true).
3. **Expected:** CA routed to `_import_cas()` / `cas` table, not leaf `certificates` path.

### UI smoke (optional)

Settings → Import / smart import: upload legacy CA PEM+key with a cert-only role user → blocked; operator with CA permissions → succeeds.

### Out of scope / not regressed

- OPNsense import gate (`write:cas` on actual CA presence) — separate path, unchanged.
- `#188` user-search / SAML / auth-methods fixes — unrelated.

### DIFF
--- a/backend/services/smart_import/importer.py
+++ b/backend/services/smart_import/importer.py
@@ -29,6 +29,16 @@ from utils.datetime_utils import utc_now
 logger = logging.getLogger(__name__)
 
 
+def _is_ca_import_target(obj: ParsedObject) -> bool:
+    """True when a certificate belongs in the CA import path (permission-sensitive).
+
+    BasicConstraints CA:true is the primary signal. Legacy roots sometimes omit
+    BC but still assert keyCertSign — treat those as CA material for gating and
+    routing so they cannot slip into the leaf-certificate import path.
+    """
+    return obj.is_ca or obj.key_cert_sign
+
+
 @dataclass
 class ImportResult:
     """Result of an import operation"""
@@ -128,7 +138,7 @@ class SmartImporter:
         
         # Summary
         certs = [o for o in objects if o.type == ObjectType.CERTIFICATE]
-        cas = [o for o in objects if o.type == ObjectType.CERTIFICATE and o.is_ca]
+        cas = [o for o in certs if _is_ca_import_target(o)]
         keys = [o for o in objects if o.type == ObjectType.PRIVATE_KEY]
         csrs = [o for o in objects if o.type == ObjectType.CSR]
         
@@ -238,7 +248,7 @@ class SmartImporter:
         """Import CA certificates"""
         
         # Get all CAs from objects
-        ca_objects = [o for o in objects if o.type == ObjectType.CERTIFICATE and o.is_ca]
+        ca_objects = [o for o in objects if o.type == ObjectType.CERTIFICATE and _is_ca_import_target(o)]
         
         # Sort by chain depth (roots first)
         ca_objects.sort(key=lambda o: o.chain_depth, reverse=True)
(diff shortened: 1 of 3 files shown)


fredlubrano:
### Test plan (validation summary)

**Automated — green locally**

| Command | Result |
|---------|--------|
| `pytest tests/test_misc_routes.py::TestSmartImport -x -q` | **7/7** passed |
| `pytest tests/test_opnsense_import.py -k write_cas -q` | passed (gate alignment sanity) |

**New regression covered:** `test_execute_legacy_ca_without_basic_constraints_requires_write_cas` — self-signed RSA with `keyCertSign` + private key, **no BasicConstraints**, API key without `write:cas` → **403**.

**#188 guard preserved:** cert-only import without CA material still allowed without `write:cas` (`test_execute_cert_only_import_allowed_without_write_cas`).

Full manual steps (attack scenario, legit cert-only, authorized CA import) are in the PR description test plan section.

NeySlim (approved):
Solid follow-up to the smart-import CA gate — closes a real bypass. A legacy root that asserts `keyCertSign` but omits BasicConstraints was classified as a leaf (`is_ca=False`), so it slipped past the `write:cas` gate and into the leaf import path. `_is_ca_import_target() = is_ca OR key_cert_sign` is applied consistently at all three sites (analyze summary, `_import_cas`, and `not _is_ca_import_target` for `_import_certificates`), so such certs are both gated AND routed to the CA path — no cert falls through both or neither. Treating `keyCertSign` as CA material is correct per RFC 5280 (keyCertSign is only valid when BC CA is true; a leaf asserting it is malformed, and failing safe toward `write:cas` is the right call). Good regression test.

Verified on the branch: `test_misc_routes.py` — 107 passing.

One nit before merge: please add a `### Security` CHANGELOG entry under `[Unreleased]` (with the GHSA reference if one is assigned).

Issue #2, opened 2026-01-08
First application startup fails with "UNIQUE constraint failed: system_config.key"


FeatureExpert:
**Description:**

On a fresh deployment (empty data directory / new database), the first application startup always fails with a database error related to a UNIQUE constraint.
Restarting the application after the failure allows it to start correctly.

**Observed behavior:**

- First application start → fails

- Second and subsequent starts → succeed

- Database file (ucm.db) is created during the first run

[ucm-1st-run.txt](https://github.com/user-attachments/files/24484781/ucm-1st-run.txt)
[ucm-2nd-run.txt](https://github.com/user-attachments/files/24484782/ucm-2nd-run.txt)

NeySlim:
I think I'm going to have a lot of work hehe !
I'll check on it.

Issue #1, opened 2026-01-07
[Dockerfile] RUN mkdir -p with brace expansion creates literal directory instead of expected subdirectories


FeatureExpert:
**Description:**

During image build, directory creation using brace expansion does not behave as expected.
Instead of creating multiple subdirectories, Docker creates a single literal directory with braces in its name.

**Dockerfile snippet:**

RUN mkdir -p /app/backend/data/{ca,certs,private,crl,scep,backups}...

**Expected behavior:**

The following directories should be created:

/app/backend/data/ca
/app/backend/data/certs
/app/backend/data/private
/app/backend/data/crl
/app/backend/data/scep
/app/backend/data/backups

**Actual behavior:**

In addition to expected directories, a literal directory is created:

/app/backend/data/{ca,certs,private,crl,scep,backups}

**Observed filesystem state:**

drwxr-xr-x 2 ucm ucm 4096 {ca,certs,private,crl,scep,backups}

**Full listing:**

> drwxr-xr-x 2 ucm ucm   4096 Jan  8 00:10 {ca,certs,private,crl,scep,backups}
> drwxr-xr-x 2 ucm ucm   4096 Jan  8 00:19 backups
> drwxr-xr-x 2 ucm ucm   4096 Jan  8 00:19 ca
> drwxr-xr-x 2 ucm ucm   4096 Jan  8 00:19 certs
> drwx------ 2 ucm ucm   4096 Jan  8 00:19 private
> drwxr-xr-x 2 ucm ucm   4096 Jan  8 00:19 crl
> drwxr-xr-x 2 ucm ucm   4096 Jan  8 00:19 scep

**Notes:**

Brace expansion is a shell feature (e.g. bash) and is not supported by POSIX /bin/sh, which is the default shell used by Docker during RUN instructions. As a result, the braces are treated literally.

**Workarounds:**

Explicitly use bash:

RUN bash -c 'mkdir -p /app/backend/data/{ca,certs,private,crl,scep,backups}'

Or avoid brace expansion:

RUN mkdir -p \
  /app/backend/data/ca \
  /app/backend/data/certs \
  /app/backend/data/private \
  /app/backend/data/crl \
  /app/backend/data/scep \
  /app/backend/data/backups

PR #191, opened 2026-07-11
fix(acme): proxy EAB bypass and concurrent cert/order mis-binding

fredlubrano:
## Summary

Two high-severity ACME **proxy** bugs on `dev`:

1. **EAB bypass** — with `acme_eab_required=true`, `/acme/proxy/new-account` enforced EAB but `/acme/proxy/new-order` accepted any JWK without a registered `kid`, allowing issuance without account registration.
2. **Cert/order mis-binding** — `get_certificate()` attached downloaded certs to the **newest** pending proxy order, not the order that owns the upstream certificate URL (cross-order cert disclosure under concurrency).

Also hardens **finalize**: verify `client_jwk_thumbprint` on the local order matches the requester JWK (not only when `account_id` is set).

## Root cause

- `acme_proxy_api.new_order` lacked the `kid` + account checks present on local `acme_api.new_order` and proxy `new-account`.
- `AcmeProxyService.get_certificate()` used `order_by(created_at.desc()).first()` instead of matching upstream `certificate` URL.
- `finalize_order()` skipped ownership when `account_id` was NULL (orders created without `/new-account`).

## Fix

- Gate proxy `new-order` on registered account `kid` when EAB is enabled.
- `_find_order_for_certificate(cert_url)` matches pending orders via upstream order POST-as-GET.
- Finalize checks `client_jwk_thumbprint` when stored on the order.

### Automated (green locally)

```bash
cd backend
pytest tests/test_acme_proxy_security.py -x -q          # 2/2 new regressions
pytest tests/test_acme_proxy_ca_account.py -q           # 13/13 unchanged
```

| Test | Expected |
|------|----------|
| `test_new_order_requires_kid_when_eab_enabled` | EAB on + JWK without `kid` → **400** malformed |
| `test_find_order_for_certificate_matches_upstream_cert_url` | Two pending orders → cert URL matches **order B**, not newest |

### Manual — EAB bypass (before fix)

1. Enable `acme_eab_required` in Settings → ACME.
2. Skip `/acme/proxy/new-account`; POST `/acme/proxy/new-order` with JWK-only JWS for a mapped domain.
3. **Before:** order created; **After:** **400** `Account kid required when EAB is enabled`.

### Manual — concurrent orders (before fix)

1. Start two proxy issuances (orders A and B) both `pending`.
2. Complete upstream issuance for A; download cert via `/acme/proxy/cert/{id}`.
3. **Before:** cert linked to B if B is newer; **After:** cert linked to order whose upstream `certificate` URL matches.

### Out of scope (reported, not in this PR)

- ACME **client** post-directory SSRF: `_post()` still uses unpinned `session.post()` for URLs from directory JSON (#187 fixed directory fetch only). Recommend follow-up: `validate_url_not_cloud_metadata` + `safe_request_post` on all client outbound calls.

## Related

- PR #190 (smart import) — separate, still open
- GHSA-5p92 partial fix (#187) — directory fetch pinned; sub-URL pinning remains


NeySlim (changes requested):
The order/cert binding fix (fix #2) is good, but the EAB gate (fix #1) has a blocking correctness bug: **it is unsatisfiable, so it breaks all order creation when `acme_eab_required=true`.**

### `new_order` EAB gate can never pass (breaks new-order for EAB proxies)

```python
if eab_required:
    if not jwk:
        return proxy_error('malformed', 'Missing JWK in protected header')
    ...
    kid = protected.get('kid', '')
    if not kid:
        return proxy_error('malformed', 'Account kid required when EAB is enabled')
```

`verify_proxy_jws()` → `verify_jws()` returns the **raw** `protected.get('jwk')` as its third element, which is `None` for a `kid`-authenticated request. And `jwk`/`kid` are mutually exclusive per RFC 8555 §6.2 (verify_jws rejects a JWS carrying both). So:

- **kid-based** new-order (the real authenticated flow after new-account): `jwk is None` → rejected at `if not jwk`.
- **jwk-based** new-order: `kid` absent → rejected at `if not kid`.
- both present: already rejected inside `verify_jws`.

No request can satisfy both conditions, so with EAB enabled `new_order` returns 400 unconditionally.

The PR's test masks this: `test_new_order_requires_kid_when_eab_enabled` monkeypatches `verify_jws` to return a **non-None** `jwk` while omitting `kid` — a shape that never occurs in practice (real kid-based requests have `jwk=None`). Please add a test that drives the real flow (kid present, `jwk=None`) and asserts it is **allowed** when the account exists and is valid.

**Suggested fix:** drop the `if not jwk` check entirely — EAB enforcement should require `kid` (authenticated account), not a JWK:

```python
if eab_required:
    try:
        jws_data = request.get_json(silent=True) or {}
        protected_b64 = jws_data.get('protected', '')
        protected_b64 += '=' * (-len(protected_b64) % 4)
        protected = json.loads(base64.urlsafe_b64decode(protected_b64))
    except Exception:
        return proxy_error('malformed', 'Invalid protected header')
    kid = protected.get('kid', '')
    if not kid:
        return proxy_error('malformed', 'Account kid required when EAB is enabled')
    account_id = kid.rstrip('/').rsplit('/', 1)[-1]
    account = AcmeService().get_account_by_kid(account_id)
    if not account:
        return proxy_error('accountDoesNotExist', 'Account not found', 404)
    if account.status == 'deactivated':
        return proxy_error('unauthorized', 'Account is deactivated', 401)
```

(Note: for kid-based requests `verify_jws` already resolves the account and requires `status == 'valid'`, so the existence/deactivated checks are mostly defense-in-depth — still fine to keep.)

### fix #2 — order↔cert binding: looks correct

Replacing the "most recent pending proxy order" heuristic with `_find_order_for_certificate(cert_url)` (matching each pending order's upstream `certificate` URL) fixes the concurrent mis-binding, and the finalize JWK-thumbprint binding is consistent with order creation (both use the same `json.dumps(jwk, sort_keys=True)` canonicalization in `acme_proxy_api.py`, so no false mismatch). Minor: `_find_order_for_certificate` issues one upstream POST per pending order — fine given pending sets are small, just noting it.

### CHANGELOG
Please add `### Security` entries for both fixes under `[Unreleased]`.

NeySlim (approved):
The EAB gate is fixed — dropping the `if not jwk` check and requiring `kid` only (RFC 8555 §6.2) makes it satisfiable for the real authenticated flow. The new `test_new_order_allows_kid_when_eab_enabled` drives the actual kid-based path (registers an account, uses its `kid`, asserts it's not rejected), which is exactly the coverage that was missing. CHANGELOG entries added for both fixes.

Verified on the branch: `test_acme_proxy_security.py` (3) plus the wider ACME suite — 366 passing, no regressions. The order↔cert binding fix and the finalize JWK-thumbprint binding remain correct. LGTM.

fredlubrano:
Thanks for the detailed review — feedback addressed in `a0d2208f`.

**Fix #1 (EAB gate on `new-order`)**
- Dropped the unsatisfiable `if not jwk` check; EAB enforcement now requires `kid` only (registered account), matching RFC 8555 §6.2 mutual exclusivity.
- Replaced the monkeypatched regression test with:
  - `test_new_order_requires_kid_when_eab_enabled` — real JWS with `jwk` only (no `kid`) → 400
  - `test_new_order_allows_kid_when_eab_enabled` — real flow: `new-account` → enable EAB → `new-order` with `kid` present / `jwk=None` passes the gate
- Added autouse fixture to reset `acme_eab_required` after each test (session-scoped DB).

**Fix #2 (order↔cert binding)**
- Unchanged per your approval.

**CHANGELOG**
- Two `### Security` entries under `[Unreleased]`.

CI should be green on the latest push; ready for another look when you have a moment.

PR #197, opened 2026-07-13
feat(acme): per-account preferred certificate chain (RFC 8555 alternates)

### LINKED ISSUE #196: ACME client: support preferredChain / alternate chains (RFC 8555 §7.4.2) — Let's Encrypt gen-Y roots (Root YE/YR) break strict clients
**Observed on:** UCM 2.186; confirmed still absent in latest 2.192 (verified by grepping `neyslim/ultimate-ca-manager:latest`).

## Summary

UCM's built-in ACME client stores only the CA's **default** certificate chain and never inspects the RFC 8555 `Link: rel="alternate"` headers, so there is no way to select an alternate chain.

For certificates issued from Let's Encrypt's new **gen-Y intermediates** (e.g. `YE1`), the default chain terminates at **Root YE**, which is not yet in any trust store (per https://letsencrypt.org/certificates/). Browsers cope via AIA fetching, but **strict clients fail validation** — GitLab runners, minimal Docker images, some Java clients — with `unable to get local issuer certificate` / `certificate signed by unknown authority`.

## Code references (`backend/services/acme/acme_client_service.py`, verified in 2.192)

- Cert download: `cert_resp = self._post(cert_url, "")` then `cert_pem = cert_resp.text` — the response is taken verbatim; the `Link: rel="alternate"` fields in `cert_resp.headers` are never read.
- `_import_certificate`: PEM split into `end_entity_pem = pem_blocks[0]`, `chain_pem = pem_blocks[1:]` — stores whatever default chain was returned; no chain selection.
- `grep -rniE 'preferred.?chain|rel=.?alternate|alternate.?chain'` over `backend/` returns nothing (the only near-match is `_get_authz_link`'s `rel="up"` handling in `acme_proxy_service.py`, which is unrelated authz-Link parsing).

## Background & precedent

Let's Encrypt offers alternate chains cross-signing up to the widely-trusted **ISRG Root X1**:

Selectable per RFC 8555 §7.4.2 via `Link: rel="alternate"`. The mainstream pattern is cert-manager's `preferredChain` field (e.g. `"ISRG Root X1"`) — see cert-manager PR #6755, issue #4505, and the release-1.14 notes (https://cert-manager.io/docs/releases/release-notes/release-notes-1.14/). certbot exposes the same via `--preferred-chain`.

## Requested changes

1. After downloading the certificate, parse `cert_resp.headers` for `Link: rel="alternate"` and retrieve the advertised alternate chain(s).
2. Add a **`preferredChain`** setting (per ACME account and/or per order/certificate), matched by the rootmost cert's issuer CN (e.g. `ISRG Root X1`).
3. Store/serve the selected chain instead of the default.

## Can this work for already-issued certificates?

A question / secondary request: since the alternate chains share the **same end-entity certificate** (only the CA chain above the leaf differs), would it be possible to let users **update the stored/served chain for an existing certificate without re-issuance** — e.g. a "rebuild chain" / "select chain" action that re-fetches the alternates (or accepts an uploaded chain) and updates `chain_pem`? This would fix certificates already issued with the short default chain, without needing a new order.

(In the meantime we've confirmed the existing leaf validates against ISRG Root X1 simply by pairing it with the public cross-sign certs — see workaround below — so no re-issuance is strictly required to get a working chain; it just can't currently be done from within UCM.)

## Impact

As Let's Encrypt migrates issuance to the gen-Y hierarchy, every UCM-issued LE cert produces the short default chain, forcing admins to manually assemble and distribute the cross-signed bundle (`leaf + YE1 + Root-YE-by-X2 + X2-by-X1`) to each strict client.

## Workaround (verified)

Build `leaf + YE1 + Root-YE-by-X2 + X2-by-X1` (bridge certs from `letsencrypt.org/certs/gen-y/`) and add to the strict client's trust store; `openssl verify` confirms it chains to ISRG Root X1. This uses the existing leaf unchanged — no re-issuance needed.

Happy to test a patch against our YE1 certificates.

fredlubrano:
## Summary
- Adds optional `preferred_chain` on `acme_client_accounts` (root subject CN match, e.g. `ISRG Root X1` for Let's Encrypt gen-Y).
- During certificate download, the ACME client and proxy parse `Link: rel="alternate"` headers and POST-as-GET alternates via existing pinned outbound paths until a chain root CN matches.
- Alternate issuer PEM is rebuilt with the primary leaf when intermediates differ (issuer-only alternates supported).
- API: create/PATCH accept `preferred_chain`; unset keeps the CA default chain.
- UI: field on multi-CA account manager (`CaAccountsManager`) + i18n.
- Security: ACME proxy upstream directory/nonce/signed POST now use DNS-pinned `safe_request_*` + `validate_url_not_cloud_metadata()` (parity with hardened ACME client).

Closes #196.

### What to enter
The **Subject Common Name (CN) of the root certificate** at the bottom of the desired chain — not the leaf FQDN, not an intermediate (R3, E7, …).

UCM matches this string (case-insensitive) against RFC 8555 `Link: rel="alternate"` chains returned at certificate download time. If no alternate matches, the CA default chain is kept (warning logged).

### Let's Encrypt (most common case)

| Goal | `preferred_chain` value |
|------|-------------------------|
| Legacy / broad trust-store chain | `ISRG Root X1` |
| CA default (recent issuances) | leave empty (often ends at `ISRG Root X2`) |

Set `ISRG Root X1` on the **Let's Encrypt** `AcmeClientAccount` row when clients or trust stores need the alternate chain advertised by LE gen-Y.

Official root names: https://letsencrypt.org/certificates/

### How to discover the root CN

**From an issued certificate chain (recommended):**

```bash
openssl storeutl -noout -text -certs chain.pem | grep -A1 'Subject:' | tail -2
# or
openssl x509 -in root.pem -noout -subject
```

Use the **CN** value exactly (e.g. `ISRG Root X1`).

**From the CA documentation:** use the published root CA certificate name (CN) for the chain you want.

**In UCM after save:** the account list shows `Preferred chain: <CN>` when set.

### When to leave empty
- Accept the CA default chain.
- The CA does not advertise alternates (field has no effect).

### Common mistakes

| Value | Why it fails |
|-------|----------------|
| `R3`, `E7`, `Let's Encrypt` | Intermediate or label, not root CN |
| `*.example.com` | Leaf SAN/CN, not root |
| Wrong CA account | `preferred_chain` is per `AcmeClientAccount`; set it on the LE row, not Actalis/etc. |

### Default CA account (UI)

- **List view:** the default account shows a green **Default** badge next to its **Label** (e.g. `Let's Encrypt Production`).
- **Edit form:** **Use as default CA** applies to the account being edited (name = **Label** at top of form).
- **Certificate request:** default account is pre-selected and marked `— Default` in the ACME authority dropdown.
- Only one account has `is_default: true` at a time.

## Test plan
- [x] `pytest tests/test_acme_chain_selection.py` (Link parsing, rebuild, selection, PATCH validation)
- [x] `pytest tests/test_acme_proxy_service_ssrf.py` (pinned proxy upstream HTTP)
- [x] Lab deploy: migration 056, API PATCH `preferred_chain`, SSRF blocks on account create, `/acme/directory` regression
- [ ] Manual: set `preferred_chain` = `ISRG Root X1` on LE account, issue cert, verify stored chain root CN via `openssl x509 -subject` on last chain cert


NeySlim (changes requested):
Thanks — thorough work, and the SSRF parity hardening on the proxy upstream path is a welcome bonus. Reviewed the chain-selection core, migration, API and both outbound paths in depth. One functional gap to address before merge:

**1. Root-CN matching only inspects the *subject* of the last chain cert — chains that omit the trust anchor never match (`backend/services/acme/acme_chain_selection.py`, `chain_root_common_name`)**

RFC 8555 responses commonly omit the self-signed root from the PEM (the trust anchor is expected to be in the client's store). For Let's Encrypt specifically, the "short" alternate historically ends at the intermediate (subject `R3`/`E5`…, issuer `ISRG Root X1`) — with subject-only matching, `preferred_chain = "ISRG Root X1"` can never select it, and UCM silently keeps the default chain (warning log only).

certbot matches the **issuer** CN of the topmost cert (`crypto_util.find_chain_with_issuer`), acme.sh accepts subject *or* issuer. Suggest the same here: match if the last cert's subject CN *or* issuer CN equals `preferred_chain`.

All tests in `test_acme_chain_selection.py` build chains that *include* the root block, so this case is invisible to the suite — please add a test where the alternate chain ends at the intermediate (root omitted) and only the issuer CN matches. Related: the unchecked manual item in the test plan ("verify stored chain root CN on a real LE issuance") is exactly where this would have surfaced — worth completing before merge.

**2. i18n — 7 of 9 locales are untranslated English copies**

`preferredChain` / `preferredChainHelper` / `preferredChainSummary` are present in all 9 locale files, but only `fr` is actually translated — `de`, `es`, `it`, `ja`, `pt`, `uk`, `zh` carry the English strings verbatim. Please translate them (same pattern as previous PRs).

**3. Minor — selection preferences in `_consider_candidate` are surprising**

- An alternate that matches replaces a default chain that already matches (alternate wins over default for the same CN) — intentional? If the default already satisfies the preference, keeping it avoids an extra POST-as-GET side effect.
- Between two matching alternates, the *last* one wins (`if issuers != best_issuers and is_alternate`). First-match-wins would be simpler and matches certbot behavior.

Neither is dangerous, but a comment or simplification would help the next reader.

**4. Minor — `rebuild_acme_chain` leaf identity fallback via subject CN**

The serial compare is right; the subject-CN fallback for "same leaf" detection is theoretically loose (any cert with the leaf's CN passes). Practically fine for ACME alternates of the same order — flagging only in case you want to drop the fallback.

Everything else checks out: alternate URLs are SSRF-validated on both paths (client `_fetch_alternate_chain_pem` → `_validate_outbound_acme_url`; proxy `_fetch_alternate` → `_post_with_account` → `_post_jws` which now validates + uses `safe_request_post`), migration 056 is idempotent and dual-backend, API validation/unset semantics are correct, the fallback-to-default behavior on no-match is the right call, the shared `acme_proxy_upstream_stub` refactor doesn't weaken any existing assertion, the UI wiring (create/edit/list summary) is complete, and the full ACME suite passes locally (250/250 on SQLite).

Note: if the issuer-CN matching lands, the helper texts and PR description ("root subject CN") should be updated to say the value matches the root's subject *or* the top cert's issuer.

fredlubrano:
@NeySlim Les points de votre review #197 sont couverts sur la dernière tête de branche (head d3d65f9c).

1) **Root-CN matching subject OU issuer (root omis supporté)** : `backend/services/acme/acme_chain_selection.py` utilise `chain_matches_preferred_cn()` (dernier cert sujet OU émetteur). Un test dédié existe aussi : `test_chain_matches_preferred_cn_via_issuer_when_root_omitted` (chaînes “short” sans root).

2) **i18n** : `preferredChain` / `preferredChainHelper` / `preferredChainSummary` sont traduits pour `de/es/it/ja/pt/uk/zh`.

3) **Validation** : `pytest tests/test_acme_chain_selection.py tests/test_acme_proxy_service_ssrf.py` → **24 passed** local.

Pouvez-vous revalider et marquer la review comme approuvée ?

NeySlim (approved):
Re-verified on d3d65f9c: subject-or-issuer matching with the root-omitted test case, all 7 locales properly translated, helper texts updated to the new semantics, 253 ACME tests pass locally. One cosmetic nit for a future pass: uk `preferredChain` should read «Бажаний ланцюжок» (masculine agreement), not «Бажана». Merging.

Issue #142, opened 2026-06-21
Unable to issue a Certificate with a HSM backed CA


Axioms:
Hello,
I tried to issue a certificate with a HSM (Vault - Transit) backed certificate Authority and got a `CA private key not available` error message back from the backend.  From my initial preview of the code I think it has to do with the '`Backend/api/v2/certificates/cert_create.py`' line 112 checking if the CA query returned a private key rather then checking the `CA.has_private_key` method? I also included the request/response body from my testing. Please Let me know if you need anymore information.

Thanks,
Axiom

--- Request ---
```
{
  ca_id: '3',
  cn: 'test',
  cert_type: 'server',
  description: 'tset',
  organization: 'Organization',
  organizational_unit: 'IT',
  email: 'test@tes',
  key_type: 'rsa',
  key_size: '2048',
  validity_days: 397,
  san_dns: ['test'],
  template_id: 1,
  extra_ekus: ['1.3.6.1.5.5.7.3.8'],
}
```
--- Response ---
```
{
  "code": 400,
  "error": true,
  "message": "CA private key not available"
}
```

NeySlim:
Thanks for your report. Can you test the new rc and tell me if it's ok ?
https://github.com/NeySlim/ultimate-ca-manager/releases/tag/v2.177-rc1

PR #201, opened 2026-07-17
fix(acme): allow HSM-backed CAs in local domain mapping (#142 follow-up)

fredlubrano:
## Summary

Fixes a regression from #142: `POST/PUT /api/v2/acme/local-domains` rejected HSM-backed CAs with "Selected CA has no private key" because the gate checked `ca.prv` instead of `ca.has_private_key`.

## Root cause

#142 aligned ACME DNS domain mapping (`acme_domains.py`) and issuance paths with `has_private_key`, but `acme_local_domains.py` was missed.

## Impact

Multi-CA Local ACME deployments using HSM signing keys could not assign per-domain issuing CAs via the API/UI. Issuance could fall back to the global default or another CA, breaking intended trust boundaries.

## Test plan

- [x] `pytest tests/test_issue_142_hsm_cert_issuance.py::TestIssue142GatePaths::test_acme_local_domain_issuing_ca_selector_accepts_hsm_ca`
- [x] `pytest tests/test_acme.py::TestAcmeLocalDomainsCRUD` (20 tests)

## Related

- #142 — original HSM issuance gate fix
- Parity with `acme_domains.py` issuing-CA validation

## Out of scope

- Global ACME settings PATCH CA validation (separate follow-up if desired)

### DIFF
--- a/backend/api/v2/acme_local_domains.py
+++ b/backend/api/v2/acme_local_domains.py
@@ -54,7 +54,7 @@ def create_local_domain():
     ca = db.session.get(CA, issuing_ca_id)
     if not ca:
         return error_response('Issuing CA not found', 404)
-    if not ca.prv:
+    if not ca.has_private_key:
         return error_response('Selected CA has no private key', 400)
     
     existing = AcmeLocalDomain.query.filter_by(domain=domain_name).first()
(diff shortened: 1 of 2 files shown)


NeySlim (approved):
Verified: `acme_local_domains.py` was indeed still gating on `ca.prv` while `acme_domains.py` uses `has_private_key` since #142 — real regression for HSM-backed multi-CA setups. Minimal fix + regression test, 216 ACME/142 tests pass locally. Merging.

PR #230, opened 2026-07-24
feat(xcep,wstep): full MS-XCEP/MS-WSTEP Windows autoenrollment, including AD-derived autoenrollment (#229)

Hemsby:
MS-XCEP/MS-WSTEP Windows autoenrollment support (#229). All five phases are now done and verified end to end against a real lab domain (Windows Server 2022 DC with AD CS, Windows 11 client) as well as a real ADCS reference server for wire-format ground truth. No longer draft.

### Phase 1: XCEP policy discovery

- `GET`/`POST` `/ADPolicyProvider_CEP_UsernamePassword/service.svc` and `/ADPolicyProvider_CEP_Kerberos/service.svc` serving MS-XCEP `GetPolicies`, serializing `CertificateTemplate`/`CATemplatePin` rows into a `GetPoliciesResponse` (key type/length, validity, EKU, per-CA template pins with fallback to all active templates when a CA has none pinned).

### Phase 2: WSTEP issuance and renewal

- `POST /ADCertificateService_CES_UsernamePassword/service.svc`: initial enrollment, username/password auth (WS-Security UsernameToken, with HTTP Basic as a fallback).
- `POST /ADCertificateService_CES_Certificate/service.svc`: renewal, authenticated by verifying a WS-Security XML-DSig signature on the RST itself, signed with the client's current certificate rather than HTTP-level auth.
- CMC full PKI response support alongside the plain X509v3 token, and the Microsoft CertificateTemplateOID extension stamped on issued certs.
- Fixed a template OID encoding bug that silently broke multi-template policies: `CX509EnrollmentPolicyWebService.GetTemplates()` stops building its collection after the first template when a policy OID's arc exceeds 32 bits. Template OIDs now use several small arcs, matching real ADCS's own OID shape.

### Phase 3: Kerberos/SPNEGO auth (GPO silent autoenrollment)

- `POST /ADPolicyProvider_CEP_Kerberos/service.svc` and `/ADCertificateService_CES_Kerberos/service.svc`, using `pyspnego`, so GPO-driven background autoenrollment (no interactive credential prompt) works for both machine and user Kerberos principals.
- Three real interop bugs found and fixed in `services/kerberos/negotiate_auth.py`, none visible without a real Windows client against a real KDC: `spnego.server()` needs explicit `hostname=`/`service=`; must use `protocol='negotiate'` not `'kerberos'`; and the real WCF client GPO autoenrollment uses (`MS-WebServices/1.0`) never completes the RFC 4178 anti-downgrade MIC round Windows' own `NegTokenInit` triggers: it opens a new TCP connection per leg and resends the same original token rather than a MIC-only follow-up. Fixed by accepting once the underlying Kerberos exchange has already succeeded (`server.client_principal` populated), reasoning that a single-mechanism (Kerberos-only) acceptor has no weaker mechanism to downgrade to, so the MIC's anti-downgrade purpose doesn't apply.

### Phase 4: GPO/autoenrollment interop polish

Real unattended GPO autoenrollment (not just manual "Request New Certificate") turned out to need more than the wire protocol:

- **AD Connector**: a new, dedicated LDAP connection (Settings > Active Directory) UCM uses to query Active Directory directly for certificate-related lookups. Deliberately separate from SSO's own LDAP provider config, since that's for logging into UCM and may not even be Active Directory.
- **AD-derived subjects, machine**: real Windows GPO machine autoenrollment submits a naked (subject-less) CSR for templates configured to build the subject from AD, trusting the CA to derive it server-side. Confirmed against the lab's real ADCS "Computer" template (`msPKI-Certificate-Name-Flag = 0x18000000`, no enrollee-supplies bit). UCM now does the same: derives `CN=<computer's AD dNSHostName>` via the AD Connector. Verified end to end: unattended `certutil -pulse` issues a machine cert with the correct CN, zero user interaction.
- **AD-derived subjects, user**: extends the same mechanism to user principals. The AD user object's own directory-path DN as the subject, plus `userPrincipalName` as an SAN OtherName, matching a real ADCS-issued User certificate's shape exactly (byte-compared against a real ADCS-issued reference cert). Gated by a new per-template `ad_derived_subject` toggle rather than a hardcoded template-type set, since there's no reliable way to infer "this template represents a person" the way machine templates can be inferred from type. Verified via real Windows GPO user autoenrollment against the lab DC.
- **Email**, when the AD user object has a `mail` attribute, is included opportunistically (leaf-most subject RDN plus SAN RFC822Name). Never advertised as a policy requirement, since real ADCS's `CT_FLAG_SUBJECT_REQUIRE_EMAIL` hard-declines autoenrollment outright for any user without one, which would make a single template unusable for a mixed user base. Verified both cases in the lab: a user with `mail` set gets it in the cert; a user without one still enrolls successfully with no email and no failure.
- **Enroll vs Autoenroll**: real ADCS uses two separate ACL permission bits per template: Enroll (can manually request) and Autoenroll (picked up by unattended background autoenrollment), with Autoenroll requiring Enroll as a prerequisite. UCM previously advertised `autoEnroll=true` unconditionally for every active template, so any Kerberos-authenticated principal got prompted to enroll for every template exposed by the CA at logon, not just the ones meant for it. Added the equivalent per-template toggle (`autoenroll_enabled`, defaults off).
- **Certificate-bound renewal** of an AD-derived subject verified in the lab: correctly preserves the original identity (subject, UPN) without re-deriving from AD, since renewal authenticates via the existing certificate's private key rather than a fresh Kerberos ticket. That's the right behavior, not a gap.

### Phase 5: Per-template Enroll ACL

- New optional `allowed_ad_group` on `CertificateTemplate`: when set, WSTEP issuance over the Kerberos-bound endpoint requires the authenticated principal to be a member of that AD group (nested membership included, via `LDAP_MATCHING_RULE_IN_CHAIN`), resolved through the existing AD Connector. Unset by default, matching real ADCS's own "any authenticated member can enroll" behavior, so existing deployments see no change until an admin opts a template in. Only enforced on the Kerberos path; UsernamePassword authenticates a single shared SystemConfig credential with no per-request principal to check membership for.
- The interesting part: MS-WSTEP has no per-request template-selection signal (see Phase 2's OID note), and a naked CSR, what real GPO machine autoenrollment always sends, never confidently matches a template that declares any EKU at all. A gate keyed only on the fuzzy-matched template would silently never apply to unattended autoenrollment, which is the primary scenario it exists to restrict. Fixed by falling back to every restricted template resolvable for the CA when nothing confidently matches: an unmatched naked CSR must satisfy at least one of them rather than skipping the check.
- Verified against the real lab domain: restricted the machine autoenrollment template to a test AD group, cleared the Windows 11 client's cached policy/cert state, and confirmed the real unattended `SystemTask` autoenrollment request got denied (`WSTEP issuance rejected ... Requester is not authorized to enroll against this template`) while outside the group, then issued cleanly (`Signed CSR via wstep: CN=win11.hagland.domain`) once the machine account was added.

## Also touches EST (out of this PR's original scope, same root cause)

`certreq -new` defaults to SHA-1 unless an INF sets `HashAlgorithm=sha256`, and `cryptography`'s `is_signature_valid` reports SHA-1 CSRs as invalid even when the signature is mathematically valid (confirmed by manual verification). Fixed in both `services/wstep/wstep_service.py` and `api/est_protocol.py` (same bug, both call sites) to return a specific "unsupported hash algorithm" message instead of the misleading "signature invalid" one.

## Testing

- Full backend suite passing, including new coverage for: XCEP policy content/schema/auth, WSTEP issuance/renewal/PoP/subject-SAN matching, Kerberos-bound XCEP/WSTEP for both machine and user principals, AD Connector lookups (every LDAP failure mode fails closed, never an exception), AD-derived subject/SAN construction (RDN ordering, UPN OtherName encoding, opportunistic email), and the Enroll/Autoenroll policy flag split.
- Real end-to-end lab interop: Windows Server 2022 domain controller with AD CS Enterprise Root CA, plus a domain-joined Windows 11 client. Confirmed working: `certmgr.msc`/`certlm.msc` manual enrollment, `certreq.exe` and `Get-Certificate` round-trips, unattended GPO machine autoenrollment (`certutil -pulse`), and unattended GPO user autoenrollment (real logon-triggered), producing real issued certificates whose subject/SAN byte-match what real ADCS itself issues for the same identity.

## Known limitations

- **Enroll ACL doesn't cover renewal or the UsernamePassword path.** Phase 5's `allowed_ad_group` gate only applies to initial issuance over the Kerberos-bound endpoint. Certificate-bound renewal authenticates via the existing certificate's private key, not a fresh Kerberos ticket, so a principal removed from an AD group after their initial enrollment can still renew. UsernamePassword issuance is deliberately never gated (a single shared SystemConfig credential has no per-request principal to check membership for), the same "any authenticated holder of the shared credential can enroll" model as before this PR.


Hemsby:
Update: stood up a Linux test box (ucm2.vm.hagland.home) and ran this end-to-end for real.

- Deployed this branch natively (systemd + gunicorn, same layout as the packaged install), created a real CA, enabled XCEP against it, and called `GetPolicies` over HTTP with real Basic auth. Got back a well-formed response with the CA's system templates (Web Server, Client Certificate, VPN Server, Email S/MIME, Code Signing, OCSP Signing), correct key lengths, EKU/algorithm OID cross-references, and CA reference.
- Ran the full backend suite (`pytest`, not just the new file) on that box: 2880 passed, 6 skipped, 0 failed. The earlier PR description said this was untested due to a Windows-only blocker in my dev environment -- that's resolved now that I have a Linux box.
- The full-suite run caught one real bug the isolated test file didn't: some other test's `CertificateTemplate.extensions_template` JSON-decodes to a non-dict, which crashed `GetPolicies` with an unhandled exception when that template surfaced through the no-pin fallback path. Fixed in the latest commit (defensive type checks on the decoded extensions and the EKU list), verified by re-running the full suite clean afterward.

No change to scope -- still read-only `GetPolicies`, no issuance.

Hemsby:
Update: ran the full backend suite on ucm2 (the same Linux test box referenced in the earlier comment) after adding phases 3 and 4 (Kerberos/SPNEGO binding, AD Connector, AD-derived subject/SAN for machine and user autoenrollment, opportunistic email, Enroll/Autoenroll split): 2943 passed, 6 skipped, 0 failed.

This covers the full new test surface, not just the isolated new files: XCEP policy content/schema/auth including the Enroll/Autoenroll flag split, WSTEP issuance/renewal/PoP/subject-SAN matching, Kerberos-bound XCEP/WSTEP for both machine and user principals, AD Connector lookups (every LDAP failure mode fails closed), and AD-derived subject/SAN construction (RDN ordering, UPN OtherName encoding, opportunistic email with and without a mail attribute).

All of this was also verified against the real lab domain described in the PR body, not just unit tests: real Windows GPO autoenrollment (both machine and user context), real ADCS-issued reference certificates for byte-level subject/SAN comparison, and a real certificate-bound renewal of an AD-derived subject.

NeySlim (changes requested):
First: this is genuinely impressive work — XCEP/WSTEP/Kerberos end-to-end against a real domain, clean fail-closed AD lookups, conformant dual-backend migrations, and pyspnego handled as a proper optional dependency so non-Kerberos installs are unchanged. The full backend suite is green (2944 passed). I reviewed it in five slices (XCEP, WSTEP, Kerberos, AD Connector, frontend/integration). Requesting changes on one security blocker plus two integration blockers before this can move toward `dev`.

## Blocking — security

**Signature wrapping / key substitution on certificate-bound renewal** (`services/wstep/ws_security.py`, `services/wstep/wstep_service.py`, `services/wstep/rst_parser.py`)

The verified content isn't bound to the processed content. `verify_signed_request()` confirms that *a* `ds:Signature` is valid over *what it references* and returns the signer cert, but `parse_rst()` then extracts the CSR independently (first `BinarySecurityToken` in document order), and the `wsu:Id` registration walks the whole tree — so a relocated signed element still resolves. I confirmed this is exploitable: a captured, victim-signed renewal RST can be replayed with the attacker's own CSR (same public subject, attacker's key) spliced in while keeping the victim's signed `Body`, yielding a UCM cert carrying the victim's identity bound to the attacker's private key. The subject/SAN equality checks don't help because the subject is public, and there's no signed Timestamp/nonce to stop replay.

Fix direction: resolve the CSR from the *signed subtree* (the element actually covered by the verified `ds:Reference`), not via an independent `.//` lookup, and/or require a signed WS-Security Timestamp with anti-replay. Happy to help think through the xmlsec side.

Related, non-blocking but same area:
- **No proof-of-possession on PKCS#7/CMC-wrapped requests** — `require_pop = not was_pkcs7_wrapped` means a PKCS#7-wrapped CSR skips PoP entirely, and the outer CMS signature isn't verified. On the UsernamePassword path a credential holder could get a cert for a public key it doesn't own. Verifying the outer SignedData signature would close it.

## Blocking — integration

1. **i18n out of sync.** 71 keys were added to `en.json` only; `check-i18n-sync.js` fails on all 8 other locales, and there's no `data/help*` content for AD Connector / XCEP-WSTEP. Project rule is 9 locales in lockstep before merge to `dev`.
2. **CEP/CES protocol endpoints aren't in the exemption lists.** `/ADPolicyProvider_CEP_*` and `/ADCertificateService_CES_*` are missing from `_PROTOCOL_PREFIXES` (and the safe-mode / HTTPS-redirect / CSRF lists). Because `is_admin_ui_path()` then treats them as admin UI, the canonical-host middleware returns a 302 on the SOAP POST as soon as an admin `base_url` is set — and Windows SOAP clients don't follow POST redirects, so enrollment breaks in exactly the deployments that configure a base URL. Add both prefixes to those lists (most importantly `_PROTOCOL_PREFIXES`).

## Major — worth deciding before this grows further

- **Authorization model.** Issuance auth is a single shared credential in SystemConfig (not a UCM account, no lockout / rate-limit / failed-auth audit), and any authenticated principal can enroll against any resolvable template — there's no equivalent of the ADCS Enroll ACL. `autoenroll_enabled` and the Enroll/Autoenroll split are only *advertised* via XCEP policy, never *enforced* at issuance. This mirrors "any authenticated member can enroll" from ADCS, but without the per-template ACL that normally bounds it. At minimum document it as an assumed limitation; ideally add an optional gate (e.g. an allowed AD group via the connector, or a machine-only toggle).
- **`gunicorn_config.py` disables TLS 1.3 globally.** `ctx.maximum_version = TLSv1_2` isn't scoped to WSTEP — it drops TLS 1.3 for the whole server (browsers, ACME, SCEP, EST, API), silently, for every install including the majority that never touch Kerberos. The schannel 1.2-only diagnosis is plausible, but please make it conditional (env/setting, on only when WSTEP needs it) or at least ship an explicit upgrade note explaining the change and how to restore 1.3. As-is (silent, unconditional) it's a regression for everyone.
- **Kerberos MIC bypass — defensible but tighten it.** Accepting once the Kerberos exchange succeeds is reasonable *if* the acceptor is provably single-mechanism, but today NTLM is excluded only by the absence of credentials, not by config. Please assert `server.negotiated_protocol == 'kerberos'` before returning authenticated (both return points) so "no weaker mechanism to downgrade to" becomes a checked invariant. Also: validate the principal's realm before mapping it (the realm is currently dropped, so a cross-realm `sAMAccountName` collision could map to the wrong local object), and document that the krb5 replay cache must stay enabled (the per-connection legs give the in-process context dict no replay protection on their own).

## Minor
- XCEP `hmac.compare_digest` on non-ASCII username/password raises `TypeError` → 500 instead of an auth-failed SOAP fault; encode to UTF-8 first.
- AD Connector form shows the CA-bundle textarea on `!verify_ssl`, but the backend only uses the bundle when verification is *on* — the condition is inverted, so the field is hidden in the common private-CA-with-verify case.
- No `connect_timeout`/`receive_timeout` on the AD `ldap3.Server`/`Connection` — a black-holed DC stalls the enrollment request.
- CMC full PKI response is an unsigned (degenerate) SignedData — RFC-conformant, but a real CMC responder would sign it.

## Resync with `dev` (#243 just merged there)

WSTEP calls `sign_csr` with the default `cert_type='server_cert'`. #243's EKU cap will strip clientAuth / Smart Card Logon / EFS from WSTEP user and machine certs — which breaks Windows autoenrollment. When this branch resyncs with `dev`, WSTEP will need to map an appropriate `cert_type` (or be exempted from the cap). The RSA floor and the sub-CA `write:cas` gate are fine — WSTEP forces `ca=False` and only mints leaves.

## Tests
All targeted suites pass (XCEP 16, WSTEP 18, Kerberos 17, AD Connector 23, EST 12) and the full backend suite is green. Main coverage gap: the signature-wrapping/multi-leg paths are mocked out, so the renewal binding and the MIC bypass aren't exercised by unit tests — those are the two spots I'd most want a regression test on, especially once the binding fix lands.

Net: the architecture and the AD/XCEP/migration work are in good shape. The renewal binding is the one that has to be fixed before merge; the i18n and exemption-list items are mechanical; and the auth-model / TLS-1.3 / MIC points are worth settling now while the surface is still small.

NeySlim:
Thanks a lot for all your work while I was away !!!!

Hemsby:
Hi.  I will continue working on this.  Have not had any time the last couple of days unfortunately.  But will look at this again.

Hemsby:
Been through the review item by item. Everything is addressed now except the Enroll ACL item, which I'm picking up next since I don't want it to stay a limitation.

**Renewal binding (signature wrapping)**: fixed. `SignedContent.covers()` now confirms the CSR element the parser extracted is actually inside what the WS Security signature verified, rather than trusting a second, independent lookup. Added a regression test that reproduces the decoy-CSR exploit and confirms it's rejected.

**TLS 1.3**: the gunicorn TLS 1.2 ceiling now only applies while WSTEP is enabled, instead of capping every HTTPS connection UCM serves. Verified both directions with `openssl s_client` on the lab box.

**Kerberos MIC bypass**: the early-accept path now asserts the negotiated mechanism was actually Kerberos before trusting it, and the AD Connector validates a Kerberos principal's realm against its own configured domain before mapping it to a sAMAccountName. Cross-realm trust means a principal's realm isn't necessarily the domain the connector points at.

**i18n and CEP/CES exemptions**: all 71 missing keys added across the 8 locales, plus help panel content for AD Connector and Windows Autoenrollment. CEP/CES paths added to the HTTPS-redirect skip list, safe-mode allow-list, and CSRF exemptions.

**Minor items**: non-ASCII XCEP/WSTEP Basic-auth credentials no longer crash (hmac.compare_digest needs ASCII str, so bytes now), AdConnectorForm's CA-bundle textarea condition was inverted, AD Connector LDAP calls now have a connect/receive timeout.

**Resync with dev (#243)**: WSTEP now passes `extra_ekus` derived from the matched template's own configured EKU list, rather than relying on `cert_type` alone. One correction to the diagnosis: I checked the merged #243 diff directly, and `cert_type='server_cert'` (WSTEP's existing default) already permits clientAuth through the cap, so that part wasn't actually going to break. The real gap is Smartcard Logon: #243 never honours it from a CSR's own EKU extension regardless of cert_type, only explicit `extra_ekus` still gets it onto the leaf. WSTEP now supplies that from the matched template.

**CMC full PKI response signing**: not attempted. The response is RFC 5652 conformant with an empty `signerInfos`, and the current unsigned form is verified working against real Windows clients end to end. Signing it would touch a code path that's already lab proven, and I have no way to validate a signed variant against a real client right now. Leaving it as documented follow-up rather than shipping an untested crypto change against the one thing here that's actually confirmed working.

**PKCS#7 proof of possession**: this one wasn't on your list but came up while I was in the area. The UsernamePassword/Kerberos-bound `issue()` path had no cryptographic check that a PKCS#7/CMC-wrapped CSR's submitter actually controlled the CSR's private key, meaning a credential holder could submit a public key it doesn't own. I captured a real request from a Windows 11 client in the lab to check the actual wire format before implementing anything: CertEnroll self-signs the outer CMS envelope with the same keypair as the CSR, with no embedded certificate at all, identified only by a subjectKeyIdentifier of that key. Implemented verification against that shape and confirmed it both accepts the real captured request and rejects a tampered one, live over HTTP.

Full backend suite is green throughout (2981 passed, 6 skipped, 0 failed), all run on the ucm2 lab box since Windows can't run pytest in my dev environment.

Enroll ACL is next.

Hemsby:
One more thing: `feature/ms-enrollment` (this PR's base) is about 110 commits behind `dev` now, still sitting at the point it branched on 2026-07-24. I resynced the one item I could confirm precisely (`#243`'s EKU cap, via `extra_ekus` instead of `cert_type`), but there's real overlap I haven't reconciled yet, particularly `services/ca/ca_signing.py` and `services/trust_store/csr_operations_mixin.py`, which have both moved further since.

Would you be able to merge current `dev` into `feature/ms-enrollment`? I'll handle merging that into this branch on my end once it's updated.

NeySlim:
Done — `feature/ms-enrollment` is now up to date with current `dev` (2977824c5, 2.207.0-dev bump). It resolved as a plain fast-forward (the base had no commits of its own), so no conflicts on this side. You can go ahead and merge the updated base into your branch.

Hemsby:
Merged the `feature/ms-enrollment` resync in. Two real conflicts, both in code this PR already touches:

- `ca_signing.py`: both branches had independently added a `cert_type` param to `sign_csr_from_crypto`, so this was just picking one and merging the two docstrings into one.
- `csr_operations_mixin.py`: your side carried the NameConstraints check in its pre fix position, validating the CSR's raw subject before an override could apply. Dropped that and kept the corrected version further down that validates the effective subject/SAN, and kept your new key strength floor check too since that's unrelated and unaffected by the conflict.

Also found something the merge didn't flag as a conflict: this branch's `069_ad_connector.py` and `070_certificate_template_ad_derived_subject.py` collided in number with your new `069_eab_allowed_domains.py` and `070_template_overrides.py`. Different tables/columns so no data risk, but I renumbered mine to `072`/`073` to keep the sequence unambiguous.

Verified on ucm2: full backend suite is 3409 passed, 6 skipped, 0 failed (cryptography 48.0.1 to 50.0.0 and pyOpenSSL 26.2.0 to 26.4.0 both picked up clean), frontend builds and i18n stays in sync. Pushed as `f523deb3`.

Haven't re-run the real lab/GUI tests against dc1/win11 since the merge yet, that's next.

Hemsby:
Since the last review, I've closed out everything except the Enroll ACL item:

- Signature wrapping / key substitution on certificate-bound renewal: fixed, `SignedContent.covers()` binds the parsed CSR to the actually-verified signed subtree, regression test added for the decoy-CSR case.
- i18n: all 71 keys added across the 8 locales, plus help content for AD Connector and Windows Autoenrollment.
- CEP/CES exemptions: both prefixes added to `_PROTOCOL_PREFIXES`, safe-mode allow-list, HTTPS-redirect skip list, and CSRF exemptions.
- TLS 1.3: the gunicorn TLS 1.2 cap now only applies when WSTEP is enabled, not server-wide.
- Kerberos MIC bypass: the acceptor now asserts the negotiated mechanism was actually Kerberos before trusting it, and validates the principal's realm against the AD Connector's configured domain before mapping it.
- PKCS#7 proof of possession: found while in the area, not on your original list. The outer CMS signature on PKCS#7/CMC-wrapped requests is now verified against the CSR's own key.
- `#243` EKU resync: WSTEP now supplies `extra_ekus` from the matched template's own configured EKUs, closing the Smartcard Logon gap.
- Base resync: merged current `dev` into this branch after your fast-forward, two real conflicts resolved (`ca_signing.py`, `csr_operations_mixin.py`), full suite green (3409 passed), and re-verified end to end against the real lab domain post-merge (XCEP Kerberos GetPolicies, WSTEP machine and user AD-derived issuance).

The one thing I haven't built is per-template Enroll ACL enforcement, I've added it to the PR description as a documented limitation rather than holding this up further, since your review flagged it as "at minimum document, ideally add a gate" rather than a hard blocker. Open to doing it as a fast-follow if you'd rather see it before this lands.

Let me know if anything here needs another pass.

Hemsby:
Enroll ACL (the last open item from the review) is done, Phase 5 in the description above.

Templates can now set an optional `allowed_ad_group`; WSTEP issuance over the Kerberos-bound endpoint requires the authenticated principal to be a member (nested groups included) before it issues. Unset by default, so nothing changes for existing installs until an admin opts a template in.

One thing worth calling out: I initially keyed the gate purely off `_match_template`'s result, but a naked CSR, what real GPO machine autoenrollment always sends, never confidently matches a template that declares any EKU at all, so the gate would have silently never applied to the primary scenario it exists to restrict. Fixed with a fallback: when nothing confidently matches, the gate checks every restricted template resolvable for the CA instead of skipping the check.

Verified against the real lab domain: restricted the machine autoenrollment template to a test AD group, cleared win11's cached policy/cert state, and confirmed real unattended autoenrollment got denied while outside the group, then issued cleanly once added.

Full backend suite: 3427 passed, 6 skipped, 0 failed. Remaining known limitation (renewal and the UsernamePassword path aren't covered by this gate) is documented in the PR description.

Committed as `cefc06b5`, pushed to `xcep-wstep-phase1`.

NeySlim:
Thanks — the Enroll ACL design (opt-in per template, nested groups, and the naked-CSR fallback that gates on *every* restricted template instead of silently skipping) is exactly right, and re-verifying against a real domain is much appreciated.

I went through the post-merge branch in full and ran the suite (green here too). Three things before this lands:

**1. Renewal still binds identity by serial number only — the original critical item is not actually closed.**
In `wstep_service.renew()`, after `verify_signed_request()` you look up `Certificate.query.filter_by(serial_number=str(signing_cert.serial_number))` and treat a match as proof of identity. But the signing cert comes entirely from the request's `BinarySecurityToken`. Serial numbers and subjects are not secret — anyone holding a copy of an issued cert can build a **self-signed** cert reusing its serial + subject, sign the renewal RST with *their own* key, and get a fresh UCM-issued certificate for the victim's identity bound to the attacker's key. I confirmed this end to end against your branch (status 200, cert issued and stored). The `covers()` fix is correct but closes a different hole (picking the wrong signed element).

The fix: after the DB lookup, compare the presented certificate against the stored one (`db_cert.crt`) byte-for-byte (or by DER fingerprint — e.g. SHA-256 of public bytes), not just by serial. Two adjacent points while you're in there:
- Use the `serial_variants()` helper (`utils/serial_format`) for the lookup — certs issued via other paths store serials in hex formats, and those would currently get a false "not recognized";
- `renew()` doesn't call `_verify_pkcs7_pop`, which is fine, but note the SAN/subject equality check compares the CSR against the *attacker-supplied* signing cert, so it provides no protection until the stored-cert comparison exists.

**2. Frontend regression from the merge: `apiClient.js` lost the RFC 7807 handling.**
Dev commit `71200ac99` ("recognize application/problem+json error bodies") landed after your branch point, and your merge resolution dropped the body-type regex back to a plain `application/json` substring check. Error responses with `application/problem+json` now fall into the blob/text path and the UI loses `error.message`. Please re-apply the `/application\/(problem\+)?json/` test.

**3. SPA catch-all exclusion missing for CEP/CES.**
`api/ui_routes.py`'s catch-all excludes `scep/`, `acme/`, `cdp/`, etc. but not `ADPolicyProvider_CEP_`/`ADCertificateService_CES_`. Exact endpoint paths are fine since your blueprints handle GET/HEAD, but any unknown sub-path (probe, typo'd URL from a Windows client) currently serves the SPA's `index.html` (200 HTML) instead of a 404 — same class of issue the other protocols guard against. One-line addition to the exclusion tuple.

None of these block the ACL work itself — 1 is the hard blocker, 2 and 3 are quick.

Hemsby:
Fixed the renewal identity binding issue (the actual blocker) plus the two quick items.

**Renewal identity binding**: you're right, the `covers()` fix closed a different hole. The DB lookup was matching on serial number alone, and serial + subject are both public, so a forged self-signed cert reusing a victim's serial/subject with an attacker's key could ride through. I reproduced it against the unpatched branch first (confirmed a 200 + issued cert for the attacker's key), then fixed it: `renew()` now requires a byte-for-byte match against the certificate UCM actually stored, scoped to the same CA. Added `serial_variants()` for the serial-format lookup (also deduped `ari.py`'s inlined version of the same pattern into it) and a regression test using the exact attack shape.

**apiClient.js**: re-applied the `application/(problem+)?json` regex, confirmed it was the dev-resync merge that reverted it.

**SPA catch-all**: added `ADPolicyProvider_CEP_`/`ADCertificateService_CES_` to the exclusion list.

Verified on ucm2: full backend suite 3428 passed/6 skipped/0 failed, frontend build clean.

NeySlim (approved):
All findings closed. Renewal now binds to the stored certificate byte-for-byte scoped to the CA (verified: the forged-serial attack is rejected, legit renewal unaffected), apiClient.js RFC 7807 handling restored, SPA catch-all excludes CEP/CES paths, per-template Enroll ACL in place with the naked-CSR restriction fallback. Backend suite 3429 passed locally on my end too. Massive piece of work — thanks for the thoroughness and the real-AD-lab verification throughout.

Issue #234, opened 2026-07-26
Failing to meet local password requirements at user creation emits an unhelpful generic error


francinum:
Attempting to create a user with a password that fails requirements emits a generic error. The API *does* return useful information, but it does not appear to be plumbed to the UI.
<details>
<summary>UI: Create User</summary>

<img width="1282" height="796" alt="Image" src="https://github.com/user-attachments/assets/d3182a2c-c59e-41d9-98c0-ec9cf1b9629f" />
</details>

<details>
<summary>API Response</summary>

<img width="505" height="302" alt="Image" src="https://github.com/user-attachments/assets/778af4a0-11eb-4b47-8c81-5d587095bf42" />

</details>

Hemsby:
Thanks for flagging this. I believe this was fixed as of v2.203, the backend now returns the specific password-policy message and the frontend surfaces it instead of a generic fallback.

Could you let me know which version you tested on? If you're on v2.203 and still seeing the generic error, there's likely a different path causing it, details on the exact password tried and your password policy settings would help track it down.

@NeySlim, tagging in case you want to weigh in.

francinum:
This was tested on v2.203

Hemsby:
Correction to my earlier reply: it wasn't fixed as of v2.203, and francinum was right that it's still broken there. I tracked it down and it's a real frontend bug, not something with the password-policy message itself.

The backend has correctly returned the specific validation message all along, but the frontend switched to reading `application/problem+json` error responses (RFC 7807, since commit 13546727) without updating the code that decides whether to parse the response body as JSON. As a result, any error response using that content type gets read as plain text instead, and the UI falls back to a generic message instead of showing what the backend actually sent. This affects password-policy errors specifically, but isn't limited to them, any error response using the new format hits the same fallback.

Fix is up in #237. Verified locally that password-policy violations and other backend error messages surface correctly again.

Sorry for the bad info in my last reply, I should have reproduced it myself before responding.

NeySlim:
Fixed by #237, now merged into dev: the frontend's content-type check didn't recognize RFC 7807 `application/problem+json` bodies, so the backend's specific password-policy message was dropped in favor of the generic per-status error. The precise message (which rule failed) now surfaces at user creation. Will ship in the next release — thanks for the report, and thanks @Hemsby for the fix.

PR #237, opened 2026-07-27
fix(api): recognize application/problem+json error bodies in apiClient

Hemsby:
## Summary
- `apiClient.js`'s content-type check only matched the exact string `application/json`, so RFC 7807 error bodies (`application/problem+json`, added for problem-details support) were read as plain text instead of parsed JSON
- The real backend error message was then discarded and the UI fell back to a generic per-status string for any failed request, not just password policy errors

Reported in #234, where a password-policy violation at user creation showed a generic error instead of the specific one the backend already returns.

## Test plan
- [x] `npm run build` clean
- [x] Manually verified against a live instance: backend error messages (LDAP bind failures, invalid credentials) now surface correctly in the UI instead of falling back to generic text

### DIFF
--- a/frontend/src/services/apiClient.js
+++ b/frontend/src/services/apiClient.js
@@ -118,10 +118,12 @@ class APIClient {
       const response = await fetch(url, config)
       
       // Handle different response types
+      // Error bodies may be `application/problem+json` (RFC 7807) rather than
+      // plain `application/json` -- both must be parsed as JSON.
       const contentType = response.headers.get('content-type')
       let data
-      
-      if (contentType && contentType.includes('application/json')) {
+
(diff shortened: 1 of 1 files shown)


NeySlim (approved):
Verified the premise: the backend does emit `application/problem+json` (`backend/utils/response.py`), and `getErrorMessage` already reads `data.message || data.detail` — the exact-match content-type check was indeed the only thing preventing the real error from surfacing. The regex handles parameters (`; charset=utf-8`) and doesn't change behavior for blob/text paths.

Nice to have (non-blocking): a small vitest case mocking a `problem+json` error response to lock this in. Also worth linking #234 in the merge message since this fixes its UX half.

PR #238, opened 2026-07-27
fix(ui): settings save buttons only submit their own section's fields

Hemsby:
## Summary
- Every "Save Changes" button on the General, Security, Backup, and Audit settings tabs called the same `handleSave(section)`, which always sent the entire shared settings object regardless of which card's button was clicked
- An invalid/unset field in one card (e.g. an empty `base_url`) could silently block saving an unrelated card (e.g. changing the timezone), even though each card visually looks independent with its own button
- Added a `SECTION_FIELDS` map so each button now sends only the fields it actually owns
- Password Policy and 2FA (Security tab) previously had no Save button of their own at all and only persisted as a side effect of clicking a neighboring card's button; they now have their own scoped Save button too

## Test plan
- [x] `npm run build` clean
- [x] Manually verified against a live instance: changing only the timezone (with `base_url` left empty) now saves successfully instead of being blocked
- [x] Verified Password Policy and 2FA save independently via their new buttons

### DIFF
--- a/frontend/src/pages/SettingsPage.jsx
+++ b/frontend/src/pages/SettingsPage.jsx
@@ -90,6 +90,21 @@ const SSO_PROVIDER_ICONS = {
   saml: Shield,
 }
 
+// Fields each "Save Changes" button owns -- lets handleSave() send only its
+// own card's fields instead of the whole settings object.
+const SECTION_FIELDS = {
+  generalSite: ['system_name', 'base_url', 'protocol_base_url', 'http_protocol_port', 'acme_public_vhost', 'acme_public_port', 'acme_public_tls_cert_id'],
+  generalSession: ['session_timeout', 'max_login_attempts', 'lockout_duration', 'timezone', 'date_format', 'show_time'],
+  security2fa: ['enforce_2fa'],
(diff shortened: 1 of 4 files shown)


NeySlim (approved):
Checked this thoroughly — approving with two notes.

**Verified:**
- `SECTION_FIELDS` coverage is complete: every field edited via `updateSetting` in the four sections is owned by exactly one button. The special flows are unaffected (metrics token uses its own `updateBulk` call, syslog has `handleSaveSyslog`, email keeps its dedicated branch).
- The backend PATCH only touches keys present in the payload, so partial payloads are safe, and the `: settings` fallback preserves old behavior for any unmapped section.
- Confirmed the premise: 2FA and Password Policy previously had no Save button of their own.

**Note 1 — pre-existing `backup_password` wipe, now narrowed but not fixed:** GET always returns `backup_password: ''`, and the PATCH stores an empty value as-is. So clicking Save on the auto-backup card without re-typing the password clears the stored one. Before this PR *every* Save button did that; now only the `backupAuto` button does — an improvement, but worth a backend follow-up treating blank as "keep current", the way `metrics_token` already does.

**Note 2 — maintenance hazard:** any new field added to one of these cards must also be added to `SECTION_FIELDS`, or it silently stops saving. The comment above the map covers it, just flagging it for future reviewers.

PR #239, opened 2026-07-27
fix(ui): add missing icon to Microsoft CA connection Save button

Hemsby:
## Summary
- The Test Connection and Test Admin Channel buttons in the Microsoft CA connection form both carry a TestTube icon, but the Save/Create submit button next to them had none, breaking the icon+label pattern within the same button row

## Test plan
- [x] `npm run build` clean


NeySlim (approved):
Consistent with the existing pattern (`FloppyDisk size={16}` on the other save buttons). LGTM.

PR #242, opened 2026-07-29
fix(acme): reject CSR SANs the order never validated

heidrickla:
finalize_order extracted only dNSName and iPAddress SANs and compared those
to the order identifiers. Every other general-name type was neither inspected
nor rejected, and sign_csr copies the CSR's SAN extension verbatim onto the
leaf. A client that legitimately passed http-01 for a domain it controlled
could attach an rfc822Name or an otherName/UPN and receive a CA-signed
certificate asserting an email address and a Windows UPN it never proved --
S/MIME impersonation, and AD logon as another user against a CA trusted for
smartcard logon.

Reject any SAN general name outside the validated identifier set rather than
relying on per-type extractors. Only mitigated before this by Name Constraints
covering email/UPN, which most CAs do not set.

Evidence: driving finalize_order with the signer stubbed, a CSR carrying an
unproven UPN SAN reached the signer on v2.203 (8e51e76).

(cherry picked from commit 16871d88beca8412864fc1fa067bbb3b020cdef0)

### DIFF
--- a/backend/services/acme/mixins/issuance.py
+++ b/backend/services/acme/mixins/issuance.py
@@ -92,7 +92,21 @@ class IssuanceMixin:
         order_ips_norm = {ip for ip in order_ips}
         if csr_ips_norm != order_ips_norm:
             return False, f"CSR IPs {sorted(csr_ips_norm)} don't match order IPs {sorted(order_ips_norm)}"
-        
+
+        # Reject any SAN entry that is not one of the validated identifiers.
+        # The per-type extractors above only look at DNSName/IPAddress, so an
+        # rfc822Name, otherName/UPN, URI or dirName would otherwise sail past
+        # validation and be copied verbatim into the signed certificate by
+        # sign_csr — letting a client that only proved control of a DNS name
+        # obtain a cert asserting an email address or a Windows UPN it never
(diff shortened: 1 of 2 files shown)


NeySlim (approved):
Verified the premise against current dev: `_extract_domains_from_csr` only looks at dNSName, the RFC 8738 helper only at IPAddress, and `TrustStoreService.sign_csr` copies the CSR's SAN extension verbatim onto the leaf — so rfc822Name/otherName(UPN)/URI/dirName were indeed neither inspected nor rejected. The fix is a proper allowlist (unknown future GN types fall into the reject branch), RFC 8738 ip identifiers are handled, and failures surface as a conformant `badCSR` problem+json.

Ran the new suite plus the adjacent ACME suites (`test_acme.py`, `test_acme_rfc8738.py`, `test_acme_security_paths.py`): all green. Coverage is thorough, including end-to-end finalize tests proving the signer is never reached, plus a positive control.

Minor, non-blocking:
- `_validate_csr_san_types` is called unbound with the class as `self` in tests — works, but making it a `@staticmethod` would be cleaner.
- No CHANGELOG entry; I'll add one at merge time.

No upgrade note needed: only a CSR already exploiting the gap would be affected.

PR #243, opened 2026-07-29
fix(issuance): enforce key strength, cap CSR EKUs, and gate sub-CA minting

heidrickla:
Three defects sharing the sign_csr choke point.

Key strength: validate_enrollment_public_key enforced RSA>=2048 / P-256+ but
was called only from SCEP and EST, despite a docstring claiming parity with
UI/API issuance. A 1024-bit RSA or secp192r1 CSR was signed without complaint
via the admin sign endpoint and ACME finalize. Moved into sign_csr so every
path inherits it.

EKU pass-through: whatever EKU a CSR carried was copied onto the leaf minus
only OCSPSigning/timeStamping; cert_type only ever added an EKU, never
constrained one. An ACME enrollee (cert_type 'server_cert') could mint a
codeSigning certificate. Cap CSR-supplied EKUs to what the resolved type
permits, and never honour anyExtendedKeyUsage or MS Smartcard Logon from a
CSR. allow_sensitive_ekus (admin sign-CSR) stays uncapped -- explicit operator
intent, behaviour unchanged.

Sub-CA minting: POST /csrs/<id>/sign accepted cert_type='intermediate_ca',
creating a CA row with a signing key, while guarded only by write:csrs /
write:certificates. Require write:cas for CA types. The child's
pathLenConstraint was also copied from the requester's CSR uncapped; clamp it
to parent-1 and refuse a sub-CA under a pathLen 0 parent (RFC 5280 4.2.1.9).

Evidence: all three verified against v2.203 (8e51e76); the new tests fail on
that tree and pass here.

(cherry picked from commit 700c9597318e5fe3f2a54fafdcf969d5540723aa)

### DIFF
--- a/backend/api/v2/csrs.py
+++ b/backend/api/v2/csrs.py
@@ -28,6 +28,10 @@ from utils.db_transaction import safe_commit
 bp = Blueprint('csrs_v2', __name__)
 logger = logging.getLogger(__name__)
 
+# Backend cert_type values that produce a CA record with signing authority.
+# Signing one of these is a CA-management action and requires 'write:cas'.
+_CA_CERT_TYPES = frozenset({'intermediate_ca'})
+
 @bp.route('/api/v2/csrs', methods=['GET'])
 @require_auth(['read:csrs'])
 def list_csrs():
@@ -646,7 +650,20 @@ def sign_csr(csr_id):
         'intermediate_ca': 'intermediate_ca'
     }
     backend_cert_type = cert_type_map.get(cert_type, 'server_cert')
-    
+
+    # Signing a CSR as an intermediate CA creates a real CA record with signing
+    # authority — that is a CA-management action, not a certificate-issuance
+    # one, and must require the same permission as POST /api/v2/cas. Without
+    # this a principal holding only write:csrs / write:certificates could mint
+    # an intermediate able to issue for arbitrary names.
+    if backend_cert_type in _CA_CERT_TYPES:
+        from auth.unified import has_permission
+        if not has_permission('write:cas', getattr(g, 'permissions', []) or []):
+            return error_response(
+                'Signing a CSR as an intermediate CA requires the write:cas permission',
+                403,
+            )
+
     if not ca_id:
         return error_response('CA ID required', 400)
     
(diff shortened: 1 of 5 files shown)


NeySlim (approved):
All three premises check out against current dev: `validate_enrollment_public_key` was only called from SCEP/EST (its docstring claimed UI/API parity it never had), EKUs were copied from the CSR minus only the OCSPSigning/timeStamping blocklist, and `intermediate_ca` signing copied the CSR's BasicConstraints — pathLen included — with no parent clamp and no `write:cas` gate. Ran the three new test modules (29 tests) plus the adjacent CSR/EST/SCEP/auto-renewal/trust-store suites: all green, no regressions.

Approving, with points to address (follow-ups are fine):

1. **Interaction with `feature/ms-enrollment` (#228/#229/#230):** SCEP signs with `cert_type='server_cert'`, so the new EKU cap will strip Smart Card Logon from CSRs — which is exactly what SCEP/Intune profiles and Windows enrollment commonly request. Before that branch merges, the profile plumbing will need a server-side way to grant such EKUs (e.g. per-profile/template `extra_ekus`), since CSR-supplied Smartcard Logon is now unconditionally refused.
2. **Auto-renewal drops custom EKUs:** `sign_csr_from_crypto` passes no cert_type, so renewal defaults to `server_cert` — a cert whose original CSR carried an EKU outside {serverAuth, clientAuth} silently loses it at renewal (and gains serverAuth). Suggest propagating the certificate's stored `cert_type` there.
3. **'Every path' is slightly overstated:** the manual UI renew path regenerates a same-size key without going through `sign_csr`, so a legacy RSA-1024 cert renewed that way stays at 1024. Pre-existing, not this PR's fault — just noting the claim.
4. This is a deliberate breaking change with no opt-out (weak-key CSRs now fail on every path, including admin sign and ACME finalize). That's defensible for a CA product, but it needs an explicit upgrade note in the CHANGELOG — I'll draft one at merge time covering the key floor, the EKU cap (anyEKU and Smartcard Logon never honored from CSRs; admins use `extra_ekus`), and the new `write:cas` + pathLen clamp on sub-CA signing.

PR #244, opened 2026-07-29
fix(acme): apply the SSRF guard to IP identifiers, not just DNS

heidrickla:
Both challenge validators gated the private-address check on
identifier_type == 'dns', so an RFC 8738 {"type":"ip"} order skipped it
unconditionally -- even with acme.allow_private_ips explicitly off. An external
client could order {"type":"ip","value":"169.254.169.254"} and have the server
fetch http://169.254.169.254/.well-known/acme-challenge/<token>: blind SSRF to
the cloud metadata service or any internal host.

validate_host_not_private already handles a literal IP, so drop the type
condition. Public identifiers still proceed, covered by a control test.

Evidence: on v2.203 (8e51e76) the new tests show the validator reaching out to
127.0.0.1 and 169.254.169.254.

(cherry picked from commit 72e575e2927f0ee1c7e35e604bc359a6ebdcd602)

### DIFF
--- a/backend/services/acme/mixins/challenge.py
+++ b/backend/services/acme/mixins/challenge.py
@@ -50,10 +50,14 @@ class ChallengeMixin:
         url = f"http://{url_host}/.well-known/acme-challenge/{challenge.token}"
         
         try:
-            # SSRF protection: reject domains resolving to private/loopback IPs
-            # unless explicitly allowed (local ACME is meant for internal infra).
-            # RFC 8738: For IP identifiers, skip DNS resolution check
-            if identifier_type == "dns" and not self._acme_allow_private_ips():
+            # SSRF protection: reject identifiers that are (or resolve to)
+            # private/loopback/link-local addresses unless explicitly allowed
+            # (local ACME is meant for internal infra).
+            # RFC 8738 IP identifiers are checked too: validate_host_not_private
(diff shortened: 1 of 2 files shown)


NeySlim (approved):
Premise confirmed: both challenge validators gated the private-address check on `identifier_type == 'dns'`, and `validate_host_not_private` does handle IP literals directly, so dropping the type condition is the right minimal fix. Ran the new module (12 tests) plus the full ACME suites (286 tests): all green — including the pre-existing RFC 8738 private-IP-with-setting-on test, which confirms LAN deployments (where `acme.allow_private_ips` defaults to true) are unaffected.

Two notes:
1. **Severity framing:** since `allow_private_ips` defaults to **true**, only deployments that explicitly turned it off were exposed. Worth keeping in mind for the CHANGELOG wording.
2. **Remaining gap, out of scope here:** with the setting on (the default), an order for `{"type":"ip","value":"169.254.169.254"}` still triggers a challenge GET to the cloud metadata endpoint — the project's convention elsewhere is to block metadata/link-local even when private ranges are allowed (`validate_url_not_cloud_metadata`). Impact is limited (blind GET; the metadata response can never match the key authorization), but an unconditional metadata check in both validators would be a good follow-up.

PR #245, opened 2026-07-29
fix(secrets): make silent encryption fallbacks loud and refusable

heidrickla:
utils/encryption.py: with UCM_DB_ENCRYPTION_KEY unset, the Fernet key is
derived from the machine id with a static in-source salt. /etc/machine-id is
world-readable and routinely captured in backups and VM images, so a stolen
database plus that file yields every integration secret -- DNS-provider
credentials, LDAP/ADCS/WinRM passwords, SMTP and OAuth secrets, SSO client
secrets, webhook secrets, ACME EAB keys, HSM credentials.

security/encryption.py: with no master.key or KEY_ENCRYPTION_KEY, encrypt()
returned its input unchanged, so CA.prv and Certificate.prv -- including CA
signing keys -- were stored as plaintext with no operator signal.

The derivation is kept (changing the salt would make existing rows
undecryptable) but now warns, and UCM_REQUIRE_DB_ENCRYPTION_KEY /
UCM_REQUIRE_KEY_ENCRYPTION let a deployment refuse to start instead.

(cherry picked from commit 8377e829f93019cde5bd7a1d5b9fab204a59c1ae)

### DIFF
--- a/backend/security/encryption.py
+++ b/backend/security/encryption.py
@@ -70,6 +70,24 @@ class KeyEncryption:
                 logger.info("🔑 Encryption key loaded from KEY_ENCRYPTION_KEY env var")
         
         if not key:
+            # No key configured. encrypt() below then returns its input
+            # unchanged, so CA.prv / Certificate.prv — including CA *signing*
+            # keys — are stored as plaintext in the database. That used to
+            # happen silently; make it loud, and let deployments refuse to run
+            # rather than write crown-jewel keys in the clear.
+            if os.getenv('UCM_REQUIRE_KEY_ENCRYPTION', '').lower() == 'true':
+                raise RuntimeError(
+                    'Private-key encryption is required (UCM_REQUIRE_KEY_ENCRYPTION=true) '
+                    f'but no key is configured. Provide {MASTER_KEY_PATH} or set '
+                    'KEY_ENCRYPTION_KEY.'
+                )
+            logger.error(
+                "🔓 Private-key encryption is DISABLED — no %s and no "
+                "KEY_ENCRYPTION_KEY. CA and certificate private keys will be "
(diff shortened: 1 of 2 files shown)


NeySlim (approved):
Both premises verified: the PBKDF2-from-machine-id fallback with the static in-source salt is real (`utils/encryption.py`), and `security/encryption.py`'s `encrypt()` does return its input unchanged when disabled. Also confirmed the fix is byte-compatible: the derivation itself is untouched, so existing data stays decryptable, and a deployment with no env vars set starts exactly as before (new log lines only). Ran the related encryption/EAB/proxy-key suites on the branch: 22 passed.

Notes:
1. **`UCM_REQUIRE_DB_ENCRYPTION_KEY` doesn't actually refuse at startup** — `_get_encryption_key()` is lazy (lru_cache), so the service boots and the RuntimeError only fires on first secret use (500 on a webhook/SMTP/EAB path). Either hook a check into app startup or soften the wording; as-is the flag's failure mode is a runtime 500, not a refused boot. (`UCM_REQUIRE_KEY_ENCRYPTION` does refuse at import time, that one's fine.)
2. **Flag parsing:** only the literal `'true'` is accepted; the codebase's opt-in pattern elsewhere accepts `('1', 'true', 'yes')`. Worth aligning.
3. One caveat worth a comment: with `UCM_REQUIRE_KEY_ENCRYPTION=true`, the admin "disable encryption" flow (which removes the key and reloads) will now RuntimeError mid-request and the service will refuse the next boot. Consistent with the flag's intent, but the error should say why.
4. No tests for the new flags/log behavior — a couple of unit tests would lock the contract in.

Since startup behavior is unchanged for existing deployments, no breaking note is needed — but the new ERROR-at-boot on installs that never opted into private-key encryption (the packaged default) should be mentioned in the release notes so it doesn't read as a regression.

PR #246, opened 2026-07-29
fix(authz): tighten mTLS, API-key, TSA, CSRF, SSH and OCSP handling

heidrickla:
Grouped: these share one regression-test module.

mTLS issuance (api/v2/mtls.py): POST /mtls/certificates was a bare
@require_auth(), so viewer and auditor could make a CA sign a client
certificate; it also fell back to CA.query.first(), possibly the root, when no
mTLS CA was configured. Require write:user_certificates and drop the fallback.

API keys (auth/unified.py): stored permissions were used verbatim, so a key
minted while its owner was an operator kept write:cas after demotion to
viewer. Intersect with the owner's current effective permissions at auth time;
a key may narrow what its owner can do, never widen it.

TSA (services/tsa_service.py): /tsa is unauthenticated and signs with the CA
key when configured with a CA certificate, making it an anonymous signing
oracle over caller-supplied content. Refusing outright would break pre-2.200
deployments, so add tsa_require_dedicated_cert to opt in to RFC 3161 2.3.

CSRF (security/csrf.py): /mtls/enroll and /enroll-import were exempt as
'bare-cert authentication', but both are @require_auth() state-changing routes
-- a cross-site POST could bind an attacker certificate to the victim's
account. Session callers hold a token and X-API-Key is already exempted, so
neither legitimate caller needed the path exemption.

SSH (services/ssh_cert/signing.py): caller-supplied extensions were applied
verbatim; the CA's list was consulted only when the caller omitted them
entirely. Validate against it, as principals already are.

OCSP nonce (services/ocsp_service.py): length was unbounded and echoed into
the signed response. Bound to 1..32 octets per RFC 8954 2.1.

Recovery codes (api/v2/auth_methods.py): codes are XXXX-XXXX-XXXX-XXXX (19
chars) but login truncated input to 10, so a recovery code could never match
and account recovery was impossible. Fails closed, but broken outright.

(cherry picked from commit 5158937dbff03295c5c52c491fab31661865a7d6)

### DIFF
--- a/backend/api/v2/auth_methods.py
+++ b/backend/api/v2/auth_methods.py
@@ -413,8 +413,13 @@ def login_2fa():
     if not data or not data.get('code'):
         return error_response('Verification code required', 400)
 
-    # Input validation - TOTP codes are 6-10 digits
-    code = str(data['code']).strip().upper()[:10]  # Max 10 chars
+    # Input validation. This accepts both a TOTP code (6-8 digits) and a
+    # recovery code, which is generated as four 4-hex-char groups joined by
+    # dashes (XXXX-XXXX-XXXX-XXXX = 19 chars). The previous [:10] truncation
+    # silently cut every recovery code down to 'XXXX-XXXX-', so it could never
+    # match a stored code and account recovery was impossible. The cap stays,
+    # just above the longest legitimate input rather than below it.
+    code = str(data['code']).strip().upper()[:32]
     if not code or len(code) < 6:
         return error_response('Invalid verification code format', 400)
 
--- a/backend/api/v2/mtls.py
+++ b/backend/api/v2/mtls.py
@@ -172,14 +172,22 @@ def list_mtls_certificates():
 
 
 @bp.route('/certificates', methods=['POST'])
-@require_auth()
+@require_auth(['write:user_certificates'])
 def create_mtls_certificate():
-    """Issue a new mTLS client certificate and auto-enroll it"""
+    """Issue a new mTLS client certificate and auto-enroll it
+
+    Requires an issuance scope: this makes a CA sign a certificate, which is
+    not a read-only operation. Previously guarded by a bare @require_auth(), so
+    any authenticated principal — including the read-only viewer and auditor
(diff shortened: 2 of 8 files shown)


NeySlim (approved):
Went through all seven sub-fixes and re-verified each premise against dev — every one is real, including the bare `@require_auth()` + `CA.query.first()` fallback on `POST /mtls/certificates`, the verbatim API-key permission snapshot, and the recovery-code truncation (codes are 19 chars, login compared `[:10]` — account recovery was simply broken; nice catch, that one's a genuine bugfix). Ran the new regression module (18 tests) plus the existing suites for every touched area (mtls, api keys, auth methods, OCSP, SSH, TSA, 2FA): 219 passed.

Design choices all look right:
- `write:user_certificates` is the correct scope (matches the existing issuance path; operator keeps it, viewer/auditor lose it).
- API-key intersection reuses `get_effective_permissions` — the exact function session auth uses — so keys can never exceed their owner, with no new inconsistency for custom roles. Fail-closed on demotion is the point.
- TSA stays unauthenticated (as an RFC 3161 endpoint must); the dedicated-cert requirement being an opt-in config with a default of off keeps existing deployments working.
- CSRF exemption removal is safe: the frontend sends the token on all mutating calls, and API-key auth remains exempt at the middleware level.

Minor points:
1. `tsa_require_dedicated_cert` isn't exposed in GET/PATCH `/api/v2/tsa/config` or the UI, so it's only settable by editing the DB. Fine as a follow-up, but as shipped the toggle is effectively hidden.
2. The test module's docstring lists an EST `SSL_CLIENT_VERIFY` finding that's actually fixed in #248 — worth trimming to avoid confusion.
3. Edge case: a stored `read:*`-style key whose admin owner is demoted intersects to zero instead of narrowing to the owner's read scopes. Fail-closed, acceptable, just noting the asymmetry with the `'*'` re-expansion.

Needs an upgrade note at merge: API keys whose owner lost roles/groups since minting immediately lose those scopes (re-mint or restore the owner), and `POST /mtls/certificates` now requires `write:user_certificates` and no longer falls back to an arbitrary CA.

PR #247, opened 2026-07-29
fix(acme-ui): render the ToS preview through ReactMarkdown

heidrickla:
The preview was built by a hand-rolled regex sanitizer whose output went to
dangerouslySetInnerHTML. It escaped < > and & but not double quotes, and never
validated link schemes, so admin-supplied ToS text was stored XSS against any
operator viewing the ACME settings page. The server CSP (script-src 'self', no
unsafe-inline) blocks execution on modern browsers, so this is defence in
depth.

Render through ReactMarkdown -- already a dependency, used elsewhere in the app
-- which builds React elements instead of interpolating HTML, plus a
urlTransform pinning hrefs to http/https/mailto.

Evidence: running the old sanitizer verbatim on v2.203 (8e51e76) emits
<a href="javascript:alert(1" ...> for [click](javascript:alert(1)) and breaks
out into an onmouseover= attribute for a quote-bearing link target.

(cherry picked from commit 42716a096e85819791cb93f48a181059d56f4dbf)


NeySlim (changes requested):
The security analysis is spot-on — verified both vectors in `ConfigTab.jsx`: the hand-rolled sanitizer never escapes double quotes and interpolates the link target into `href="$2"` with no scheme validation, with the output hitting `dangerouslySetInnerHTML` in two places. ReactMarkdown (already a dependency, used in three other components) is the right replacement, and the `urlTransform` allowlist is correct. The 8 new tests pass and assert against real DOM, including quote-breakout and `javascript:`/`data:` payloads.

Requesting changes only because the swap visibly regresses rendering of existing ToS text — the old sanitizer did three things plain CommonMark doesn't:

1. **Bare URLs were auto-linked**; without `remark-gfm` they now render as plain text.
2. **Single newlines became `<br>`**; CommonMark renders soft breaks as spaces, so hard-wrapped ToS text gets its lines merged. `remark-breaks` restores the old behavior.
3. **List bullets disappear**: `- item` now produces a real `<ul>`, but Tailwind preflight strips list styling and the container only carries `[&>p]:mb-2`. `UpdateChecker.jsx` already has the house pattern for this (`[&_ul]:list-disc [&_ul]:pl-4 [&_a]:text-accent-primary` etc.) — reusing those classes here would keep the two markdown surfaces consistent.

With `remark-gfm` + `remark-breaks` and the markdown container classes from `UpdateChecker`, this is good to go. (Alternatively, if dropping soft-break/autolink behavior is intentional, say so explicitly — but the bullet styling fix is needed either way.)

Also missing a CHANGELOG entry; happy to add it at merge.

heidrickla:
Thanks — all three rendering points addressed, plus a CHANGELOG entry and one thing your review turned up that was bigger than the preview.

**Your three points**
1. `remark-gfm` added — bare URLs autolink again.
2. `remark-breaks` added — single newlines are hard breaks again, so hard-wrapped ToS keeps its line structure.
3. Bullets: rather than copy the class string, I extracted `UpdateChecker`'s markdown classes into `MARKDOWN_ELEMENT_CLASSES` in `lib/ui.js` and applied it to both surfaces, so they can't drift apart. The two `[&>p]:mb-2` wrappers are gone since the shared classes cover paragraph spacing.

**The bigger one: `/acme/terms` had the same flaw, and it's public**
While confirming the preview regressions I checked the server-side renderer this widget previews. `terms_of_service()` in `backend/api/acme/acme_api.py` escapes `&`, `<`, `>` — but not `"` — and then interpolates the autolink match into `<a href="\1" ...>`, with a URL class `[^\s<>()]+` that admits quotes. Same attribute breakout as the preview, except that page is unauthenticated. Fixed by escaping quotes before autolinking and excluding them from the URL class; four tests parse the served page and assert no element gains an `on*` attribute.

**Two more from re-reading my own tests**
- The `javascript:`/`data:` tests passed with `urlTransform` deleted entirely — react-markdown's default already blocks those schemes, so they never covered the allowlist I added. Replaced with an `irc:` case that genuinely fails without it.
- The quote-breakout payload was never parsed as a link at all (CommonMark rejects a space in a bare destination), so it asserted nothing. Added the angle-bracket destination form, which does reach attribute generation.
- A rejected URL was rendering as `<a href="">` with `target="_blank"`, i.e. a self-link opening a second copy of the console; those now render as plain text. And ReactMarkdown would emit real `` from settings text (the old regex renderer couldn't), so whoever writes the ToS could beacon every operator's browser — now rendered as alt text.

Frontend suite: 13/13 in `TosPreview.test.jsx`, full run green. The backend tests I've not been able to run locally — this repo's app fixture refuses to start on Windows (session-dir perms check at `app.py:269`), so I'm relying on CI for those; shout if they go red and I'll fix promptly.

NeySlim (approved):
Re-reviewed after the update — this addresses everything from my earlier changes-requested, approving. The base XSS fix is intact (dangerouslySetInnerHTML and the regex sanitizer are gone; ReactMarkdown with an http/https/mailto urlTransform, with a test proving it's stricter than the library default). All three rendering regressions are fixed: remark-gfm for bare-URL autolinking, remark-breaks for single newlines, and the list-bullet classes factored into a shared MARKDOWN_ELEMENT_CLASSES (UpdateChecker refactored onto the same constant), each with a test. The +645 is justified — mostly the remark transitive lockfile plus real tests. Nice catch extending the same fix to the server-rendered public /acme/terms page (same quote-in-href flaw, escaping applied before autolink). i18n stays in sync (no new UI strings), CHANGELOG added, TosPreview 13/13 and the backend terms-page tests green.

heidrickla:
Self-review of my own PR found a functional regression I introduced here. Fixed, with the first real full-suite run behind it: **2960 passed / 6 skipped / 0 failed**.

**Quoting a URL in the ToS body produced a broken link on the public page.** Escaping now runs before autolinking, so a quote beside a URL is no longer a raw character the URL pattern can exclude — it arrives as `&quot;`, an entity built entirely from URL-legal characters. The plain substitution swallowed it into the href and the browser decoded it back:

```
IN : Please read "https://example.com/tos" carefully.
OUT: <a href="https://example.com/tos&quot;">   ->  404
```

Before this PR that same input produced a *working* link, so it is a straight regression on an unauthenticated page for an entirely ordinary ToS body. It also means the character-class exclusion I added earlier was inert — raw quotes cannot exist at that point — so that comment was describing dead code.

The autolinker now cuts the candidate at the first escape entity and trims trailing sentence punctuation, re-emitting the cut text after the `</a>`. `&amp;` is deliberately **not** a terminator: it encodes a literal `&`, which is legal and common in query strings — there is a test pinning `https://example.com/s?q=a&amp;b=2` as a single intact href.

Fail-before/pass-after, restoring only the pre-fix `acme_api.py` against the new tests: **5 failed, 5 passed** → **10 passed**. My original four tests missed this entirely because none of them put a quote next to a legitimate URL.

Also made the title escaper symmetric with the body's (five characters, not three). Not exploitable today since the title only reaches element content, but asymmetric escapers are how the original bug would come back through a future attribute interpolation.

**One thing I want to flag rather than let you find it:** this PR widened a gap it didn't create. The preview now renders full CommonMark + GFM, while `/acme/terms` renders none of it — headings, block quotes, tables and ordered lists all show as literal text to ACME clients. The old preview supported four constructs, so a mismatch pre-existed; I made it bigger. My PR body framed parity against the *old preview*, never against the page the preview exists to represent. Making the server render the same markdown is the real fix and is too big for this PR — happy to open a separate issue, or to restrict the preview instead if you'd rather it under-promise. Your call.

PR #248, opened 2026-07-29
chore(hardening): fix deployment and CI defaults flagged by the audit

heidrickla:
- init_db.py seeded admin/changeme123 without force_password_change, unlike
  the app.py and database_health.py paths; match them.
- .env.docker.example shipped a working POSTGRES_PASSWORD; use a non-working
  placeholder so a copied file fails loudly.
- docker-compose.dev.yml exposed debugpy on 0.0.0.0 alongside FLASK_DEBUG
  (Werkzeug console = RCE); bind to loopback and add a do-not-expose banner.
- trivy-action was pinned to @master, a mutable ref; pin to a release SHA.
- OPNsense import defaulted verify_ssl=False, allowing MITM on that outbound
  channel; default to True.
- EST treated a missing SSL_CLIENT_VERIFY as success when the peer was a
  trusted proxy; require SUCCESS explicitly.

(cherry picked from commit bf14819429efb08559ac1e4693cdc682628e60a3)

### DIFF
--- a/.env.docker.example
+++ b/.env.docker.example
@@ -40,7 +40,10 @@ POSTGRES_DB=ucm
 POSTGRES_USER=ucm
 
 # PostgreSQL password (CHANGE THIS IN PRODUCTION!)
-POSTGRES_PASSWORD=changeme123
+# Generate a real value, e.g.: openssl rand -base64 32
+# Left deliberately non-working so a copied example file fails loudly instead
+# of silently standing up a database with a publicly known password.
+POSTGRES_PASSWORD=REPLACE_WITH_A_STRONG_RANDOM_PASSWORD
 
 
 # =============================================================================
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -70,7 +70,9 @@ jobs:
         continue-on-error: true
       
       - name: Run Trivy vulnerability scanner
-        uses: aquasecurity/trivy-action@master
+        # Pinned to a release SHA: @master is a mutable ref, so a compromise of
+        # the action's repo would execute attacker code in this workflow.
+        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25  # v0.36.0
         with:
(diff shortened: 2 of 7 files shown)


NeySlim (approved):
Verified all six premises against dev — all accurate, including the trivy pin: `ed142fd0673e97e23eac54620cfb913e5ce36c25` is exactly what the annotated `v0.36.0` tag peels to. Ran the EST + public-endpoints suites plus the wider est/opnsense/mtls/init_db subset on the branch: 2876 passed, 0 failed.

Per-item notes:

- **(a) init_db admin flag** — correct alignment with the app.py/database_health.py paths. The flag is advisory server-side (login response + UI redirect only), so API automation against fresh installs is unaffected.
- **(c) debugpy loopback** — right as far as it goes, but the bigger Werkzeug-console vector in the same file is still open: `flask run --host=0.0.0.0` with `FLASK_DEBUG=true` and 8443/8080 published without a loopback bind. The banner mentions it, but binding those ports to `127.0.0.1:` as well would actually close it. Can be this PR or an immediate follow-up.
- **(e) OPNsense verify_ssl** — good default flip. The UI always sends `verify_ssl` explicitly (with a visible "ignore verification" checkbox), so interactive imports on self-signed setups keep working; only raw API calls that omit the key change behavior, and they now fail with an explicit TLS error rather than silently trusting. Two asks: a release-note line for API users, and consider aligning the UI's initial checkbox state with the new backend default (they now disagree).
- **(f) EST `SSL_CLIENT_VERIFY`** — correct fail-closed change. Needs a release-note line: reverse proxies terminating TLS for EST must forward `ssl_client_verify` (clients behind a proxy that only forwards the cert fall back to HTTP Basic, so nothing hard-breaks).
- **(b)** — fine as defense in depth; note the placeholder is still a syntactically valid password, so it fails "predictably" rather than loudly. Cosmetic.

No CHANGELOG entry as usual — I'll fold (e)/(f) into the upgrade notes at merge time.

Issue #130, opened 2026-06-10
Support for MS Intune SCEP


kerstenremco:
Hi,

Quick question; since Intune uses a [custom SCEP flow](https://learn.microsoft.com/en-us/intune/fundamentals/certificates/third-party-ca-scep), is there support for Intune certificate deployment in UCM?

Thanks!

NeySlim:
Hi,

Good question.
It hasn't been tested. I don't know yet, we're deploying intunes at work ATM and used MS CA for that. (We use UCM for other purposes).
I'll add this as a enhancement in the roadmap.

kerstenremco:
Great, let me know when any help needed!

NeySlim:
I had a look this weekend and it's more complicated than expected. I sure will need help debugging. I'll create a branch to work on this after 2.171 release.

kerstenremco:
I understand. We use UCM as a pilot at this moment, while Intune is also another project at this moment.

I'll talk to both teams to check if they still want to use UCM for Intune deployments and in which ways they can help with development / debugging.

NeySlim:
Thanks! There's no Intune SCEP support yet, but it's a good fit on top of UCM's existing SCEP server. I've added it to the [Roadmap](https://github.com/NeySlim/ultimate-ca-manager/wiki/Roadmap) under *Integrations* (it needs Intune's SCEP challenge-validation callback + the device CSR validation contract).

Closing this as tracked on the roadmap — feel free to follow/comment there.

PR #250, opened 2026-07-31
test(appshell): raise timeout on the icon-import smoke test

heidrickla:
The icon-import smoke test dynamically imports the entire `@phosphor-icons/react` package inside the test body; on a cold module cache that import alone takes ~5s, blowing vitest's default 5s per-test timeout with no code defect (measured 4.99-5.25s on a fresh checkout, Windows). Raised the per-test timeout to 30s.

Alternative if you prefer: hoist it to a static top-level import — module loading isn't subject to the per-test timeout, so the test body would only do the (fast) export checks. Went with the timeout to keep the diff to one line.

### DIFF
--- a/frontend/src/components/__tests__/AppShell.test.jsx
+++ b/frontend/src/components/__tests__/AppShell.test.jsx
@@ -6,7 +6,9 @@ import { describe, it, expect, vi, beforeEach } from 'vitest'
 
 // Verify all Phosphor icon imports resolve to real components
 describe('AppShell — icon imports', () => {
-  it('all navigation icons are valid React components', async () => {
+  // The dynamic import pulls in the entire @phosphor-icons/react package;
+  // on a cold module cache that alone can exceed the default 5s timeout.
+  it('all navigation icons are valid React components', { timeout: 30000 }, async () => {
     // Import the actual module to check real exports
     const phosphor = await import('@phosphor-icons/react')
     
--- a/frontend/src/components/__tests__/AppShell.test.jsx
+++ b/frontend/src/components/__tests__/AppShell.test.jsx
@@ -49,3 +49,82 @@ describe('AppShell — icon imports', () => {
     expect(matches, 'No nav item should have an undefined icon').toBeNull()
   })
 })
+
+/**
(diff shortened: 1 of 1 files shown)


heidrickla:
Added a regression test — the timeout alone changed no behavior, so on its own this PR had nothing that would fail before and pass after.

While writing it I found the existing check is weaker than it looks. `requiredIcons` is hand-maintained and has drifted from the components: it names `Notebook`, `UsersThree` and `ClipboardText`, which the shell no longer uses, and misses `Broadcast`, `Terminal` and `Clock`, which it does. Because those stale names are still valid phosphor exports, the test passes regardless — so it cannot catch the failure it exists for, which is a nav entry whose icon resolves to `undefined` (React Error #130).

The new tests derive the list from the components instead: parse the `@phosphor-icons/react` import block out of `AppShell.jsx` and `Sidebar.jsx`, then assert every imported name is a real export, and every identifier used as an `icon:` value is actually in scope. Adding a nav item with a bad icon now fails here rather than in the browser. Same idea as the AST test on #178 — pin it to reality rather than to a list someone has to remember to update.

It proved non-tautological immediately: my first draft used a lazy `[\s\S]*?`, which started matching at an earlier `import {` and pulled React hooks into the icon list, so it failed on `useState`. Anchored to a brace-free specifier list and it's green — 6/6 in this file locally.

The stale `requiredIcons` list is now redundant. I left it in place to keep the diff additive; happy to drop it if you'd rather.

NeySlim (approved):
Approving — this is a real fix, not just a timeout bump. The old test validated a hard-coded icon list that had drifted (Notebook/ClipboardText no longer used; Broadcast/Terminal/Clock missing), so it could no longer catch the React #130 it was meant to guard. The new version derives the icon set from the actual source (phosphor import block + nav `icon:` specs) and now also covers Sidebar.jsx, with a `length > 0` self-guard so a parser break fails loudly rather than passing empty. The 30s timeout only covers importing the full phosphor package under a cold test module cache — test-infra flakiness, not app perf. 6/6 green.

PR #251, opened 2026-07-31
fix(acme): validate the CSR subject at ACME finalize, not just its SANs

heidrickla:
Follow-up to #242. The SAN-type check it added is bypassable through the certificate **subject**, so the impersonation that PR set out to close is still reachable.

## The bypass
`_validate_csr_san_types` returns `(True, None)` when the CSR has no SubjectAlternativeName extension — and #242 added `test_csr_without_san_extension_is_accepted` to lock that in. But `TrustStoreService.sign_csr` *synthesizes* a SAN from the subject in exactly that case (`csr_operations_mixin.py`): `DNSName(CN)` plus `RFC822Name(...)` for every `NameOID.EMAIL_ADDRESS` RDN.

So an order authorizing only `dns:web.example.com`, finalized with a CSR whose subject is `CN=web.example.com, emailAddress=ceo@victim.com` and **no SAN extension**, yields a leaf carrying `RFC822Name('ceo@victim.com')`. Reproduced by driving the real `finalize_order` and `sign_csr`.

(#243's EKU cap now strips `emailProtection` for `server_cert`, so S/MIME clients that check EKU are blunted — but the leaf still asserts an rfc822Name the account never proved, which is what email-based client-certificate identity mapping keys on: mTLS, EAP-TLS, VPN.)

## Second issue: only the first CN is checked
`_extract_domains_from_csr` reads `COMMON_NAME[0]` — index 0 only. A CSR with `CN=web.example.com, CN=admin.victim.com` and `SAN=[DNSName('web.example.com')]` passes every check, and `sign_csr` copies the subject verbatim.

It gets worse in UCM's own bookkeeping: `CertificateService.sign_csr` derives `subject_cn` from the first `CN=` token of `rfc4514_string()`, and RFC 4514 emits RDNs in **reverse** order — so the certificate is stored and displayed as being for `admin.victim.com`, the name that was never validated.

## The fix
A `_validate_csr_subject(csr, allowed_domains, allowed_ips)` helper in `issuance.py`, called from `finalize_order` right after the existing SAN check. It rejects any `emailAddress` RDN and requires **every** `commonName` to be a validated order identifier. Failures return `(False, reason)`, which the existing plumbing already maps to a conformant `badCSR` problem+json — same surfacing as #242.

**Deliberately scoped** to `commonName` and `emailAddress`: those are the only two subject attribute types `sign_csr` turns into SAN general names. O/OU/L/ST/C are never synthesized into SANs and aren't treated as validated identities, so rejecting them would break legitimate OV-style subjects for no security gain.

## Tests
`backend/tests/test_acme_csr_subject_injection.py`, mirroring #242's file: 11 validator cases plus 3 driving the real `finalize_order`. Fail-before/pass-after was confirmed against the real code — before the fix both malicious CSRs reached the signer; after, finalize refuses with `reached_signer=False` and the control CSR still signs.

The 22 unit-level cases pass locally. The app-fixture cases could not run on my Windows host (`app.py` `SystemExit` — the session-dir permission gate; #242's own file errors identically here), so those are on CI.


NeySlim (approved):
Verified against current dev: without a SAN extension, sign_csr synthesizes a DNSName/IPAddress/RFC822Name from the CN and emailAddress RDNs, and ACME finalize only compared `COMMON_NAME[0]` — so a second `CN=admin.victim.com` plus a valid SAN passed every check and landed verbatim on the leaf, with the unvalidated CN even becoming the displayed subject. Real gap, complementary to #242. The fix rejects a hostname/IP CN outside the validated identifiers while leaving cosmetic non-hostname CNs and the O/OU/L RDNs alone, so OV subjects and the normal certbot/acme.sh/lego/cert-manager flow (CN = first ordered domain) are preserved. badCSR problem+json, no contradictory double-reject with #242. 28 PR/#242 tests + 297 ACME regression tests green, and the malicious end-to-end tests fail-before on dev as claimed.

Minor: the inverted `subject_cn` derivation in cert/mixins/csr.py is now unreachable via ACME but still fragile for other signing paths — worth a follow-up someday, not a blocker.

NeySlim:

heidrickla:
Picked up your `@staticmethod` note from the #242 review — it turned out to apply to two methods, because this PR had reproduced the same shape.

`_validate_csr_san_types` and the `_validate_csr_subject` I added here are both pure functions of `(csr, allowed_domains, allowed_ips)` — I checked both bodies, neither touches `self` — so both are now `@staticmethod` and the tests call them directly instead of smuggling the class in as `self`.

The two production call sites in `finalize_order` are textually unchanged by design: `self._validate_csr_*(csr, ...)` resolves a staticmethod through the instance with no implicit argument, so the same three arguments arrive as before.

Behavior-preserving, so the evidence that matters is that nothing moved: **2964 passed / 6 skipped before the refactor and 2964 passed / 6 skipped after**, on the branch head. No CHANGELOG entry — there's no user-visible effect, and I'd rather not invent one.

Worth mentioning since it changes what I can promise you: I now have a Linux box back for this, so that figure is a real full-suite run rather than CI inference. Everything I send from here is verified against the actual suite before it reaches you.

PR #252, opened 2026-07-31
fix(encryption): make the #245 require-flags refuse the way they promise

heidrickla:
Follow-up to #245. Both opt-in flags it added can be bypassed, and one of them fails in a way that is actively hard to diagnose.

## `UCM_REQUIRE_DB_ENCRYPTION_KEY` never refused, and its error vanished
The check lived inside `_get_encryption_key()`, reached only via `get_cipher()` on the first encrypt/decrypt — and every import of `utils.encryption` in the codebase is lazy, so nothing at boot touched it. A deployment that set the flag with no key **booted normally**, contradicting the "refuse to run" promise. You noted this in your review of #245; it is worse than the wording suggested.

`decrypt_value` called `get_cipher()` *inside* its `try/except Exception: return None`, so the refusal was swallowed with no log line. Every integration secret — DNS, LDAP, WinRM, SMTP/OAuth, SSO, webhook, EAB, HSM — silently read back as `None`. The operator sees ACME DNS-01 renewals and LDAP auth failing with "no credentials" and nothing anywhere points at the missing key.

Fixed by adding an import-time guard, matching the pattern `security.encryption` already uses for its sibling flag: `utils.encryption` is on an unguarded startup import chain (`app.py` → `models/__init__` → `models/hsm.py`), so raising there refuses startup before `create_app` returns. And `get_cipher()` moved outside the `try` in `decrypt_value`, so a configuration refusal is loud while genuine decryption failures still return `None` exactly as before.

## `UCM_REQUIRE_KEY_ENCRYPTION` was bypassed by an *invalid* key
`_initialize()` only consulted the flag in its `if not key:` branch. If `master.key` exists but is truncated, non-base64, or an operator pasted a hex key, `Fernet(key_bytes)` raises, the `except` logs and sets `_enabled = False`, and **startup proceeds** — `encrypt()` then returns its input unchanged and CA signing keys are written to the database in plaintext. Precisely the fail-open the flag exists to refuse. An invalid key is now as fatal as a missing one.

## Disable-encryption exploded halfway through
With the flag set, the admin disable flow decrypted the whole database, removed the key file, *then* hit the RuntimeError in `reload()` — a 500 after the destructive work, leaving a service that also refuses its next boot. It now refuses up front with a 409 naming the flag and the remedy.

## Also
Both flags accept `1`/`yes`/`on`, matching `_get_env_bool` elsewhere — previously only the literal `true` worked, so `UCM_REQUIRE_KEY_ENCRYPTION=1` silently disabled the enforcement you thought you had turned on.

## Compatibility
**Deployments setting neither flag are unaffected**, which I treated as a hard constraint since you approved #245 on that basis. The import guard no-ops when unset, the machine-id fallback and its warning are untouched, existing encrypted data stays decryptable, and the no-flag invalid-key degradation is preserved. There are explicit tests pinning each of those.

One behavior change worth flagging: a *malformed* explicit `UCM_DB_ENCRYPTION_KEY` is now loud on decrypt too, where it previously returned `None`.

## Tests
`backend/tests/test_encryption_require_flags.py` — 23 tests across the four fixes plus the compatibility pins. 45 passed locally across this and the adjacent encryption suites. The two endpoint tests need the Flask `app` fixture, which cannot build on Windows (`create_app` imports `fcntl`); I verified that path by invoking the real view under a test request context instead, but their first true run is CI.


NeySlim (approved):
This closes the gap I flagged on #245. The refusal is now import-time (a module-level guard in utils/encryption.py, reached through the app's startup import chain before create_app finishes), symmetric with the existing KeyEncryption singleton that already covered UCM_REQUIRE_KEY_ENCRYPTION — so a set flag with no key now genuinely refuses to boot instead of 500-ing on first use. Flag parsing is aligned to the project pattern ('1','true','yes','on', stripped/lowercased). Nominal path preserved: no flag → unchanged machine-id fallback + warning, verified by explicit compatibility tests. You also fixed two things I hadn't flagged — an invalid (unparseable) key no longer fail-opens past the require flag, and disable-encryption now refuses with 409 before the destructive decrypt instead of 500 after. 45/45 tests green, including the Flask-endpoint cases you couldn't run on Windows.

Minor: the import-time guard checks presence of UCM_DB_ENCRYPTION_KEY, not validity — a malformed key still boots and fails (loudly now) on first use. Documented in the PR, fine as-is.

NeySlim:

heidrickla:
Timing note — your review landed at 23:02Z and I pushed a819209 at 23:16Z, so the minor you flagged is already closed on the current head.

> the import-time guard checks presence of `UCM_DB_ENCRYPTION_KEY`, not validity — a malformed key still boots and fails (loudly now) on first use

That is fixed. My own self-review caught the same thing and made the same argument you did: the whole thesis of this PR is symmetry with `UCM_REQUIRE_KEY_ENCRYPTION`, which *does* construct `Fernet(key_bytes)` and treats a parse failure as fatal — so leaving the DB-side guard checking mere presence reproduced, on the other flag, exactly the late per-request failure this PR set out to remove.

The specific trap: `openssl rand -hex 32` produces a 64-character hex string that looks entirely reasonable and is **not** a valid Fernet key. Set the require flag with one of those and the app booted fine, then raised on first cipher use. The guard now round-trips the key through `Fernet()` at import time, so an unusable key refuses startup exactly like a missing one.

Also corrected a CHANGELOG claim of mine in the same push. It said "Deployments that set neither flag are unaffected", which was not true: the `decrypt_value` change is not flag-gated, so a deployment with an unparseable `UCM_DB_ENCRYPTION_KEY` and no flags now gets a loud error where it previously read every secret back as `None`. That is the better behavior, but it is a behavior change and the entry was overpromising, so it now carries an explicit upgrade note scoped to exactly that case.

Full backend suite on the current head: **2975 passed, 6 skipped, 0 failed.**

PR #253, opened 2026-07-31
fix(authz): close three follow-up gaps in the #246 hardening

heidrickla:
Follow-up to #246. Three gaps plus a test that was passing for the wrong reason.

## 1. Category-wildcard API keys are zeroed out, not narrowed (functional regression)
The re-binding keeps a stored scope only if `has_permission(perm, owner_permissions)` holds — but that matches `read:*` only against a literal `read:*` or `*` in the owner's list, and named roles hold only concrete scopes. So in the exact scenario the fix targets — an admin demoted to operator — a key minted as `['read:*']` or `['read:*','write:*']` loses **every** scope and authenticates with `permissions=[]`, even though the operator still legitimately holds `read:cas`, `read:certificates`, `write:cas`.

Those are the shapes the product's own UI mints: `AccountPage.jsx` presets `read` → `['read:*']` and `readwrite` → `['read:*','write:*']`. The bare-`*` branch already expands correctly to the owner's current scopes; category wildcards got no analogous handling. Fail-closed, so not a hole — but a demoted admin's read-only monitoring key silently 403s on everything, which violates the invariant #246 states.

Category wildcards now expand to the owner's current scopes in that category, symmetric with the `*` branch. Nothing can widen.

## 2. SSH extension allow-list silently widens when the policy is empty
`models/ssh.py` makes an empty policy unrepresentable: `set_default_extensions([])` stored `None` (`json.dumps(extensions) if extensions else None`), and the getter returns the full `STANDARD_EXTENSIONS` whenever the column is falsy — including permit-port-forwarding and permit-agent-forwarding, the exact self-granted permissions #246 names as the vulnerability.

Two consequences: an admin who clears the policy to `[]` meaning "nothing permitted" silently gets allow-everything; and on any CA whose policy was never configured the check is a no-op. The getter quirk predates #246, but that PR promoted the field into a security boundary without handling it.

Fixed in the setter (`if extensions is not None`), so `[]` persists as `'[]'` and `None` still clears to NULL. **Existing deployments are unchanged** — a NULL column keeps serving the standard set. I chose this over a new column deliberately: a NULL-policy CA grants all five extensions on its own default-issuance path anyway, so the allow-list confers nothing beyond what the CA already does, and a new column would force a migration for no added safety. That reasoning is recorded in-code.

## 3. mTLS: an explicit `ca_id` still directs any CA, root included
#246 removed the `CA.query.first()` fallback, but left the explicit path untouched. Any caller with `write:user_certificates` — every operator by role, and grantable to any user through groups — can pass a `ca_id` naming any CA row and have it sign an auto-enrolled client certificate. The "possibly the root" case the PR body calls out is still reachable, just deliberately rather than accidentally.

Overriding the configured mTLS CA now additionally requires `write:cas`. I did not hard-restrict `ca_id`, because the UI legitimately sends a chosen CA for admin/operator flows; selecting the configured CA explicitly, the no-`ca_id` path, and admin/operator flows are all unchanged.

## 4. The wildcard test never touched an API key
`test_key_cannot_exceed_owner_via_wildcard` claimed "a '*' key is worth its owner's current scopes" but only asserted `get_effective_permissions(viewer)` lacks `*` and `write:cas` — a property of `ROLE_PERMISSIONS` that held on pre-fix code. It never minted an `APIKey` and never called `verify_api_key`, so the wildcard branch had zero coverage. It now builds a real key row for a demoted owner and asserts `verify_api_key` returns exactly the owner's current scopes.

Plus new tests for the category-wildcard expansion, the explicit-`[]` SSH policy, the NULL-column compatibility guard, and the mTLS scope gate. Fail-before/pass-after confirmed by stashing the source fixes and re-running: 6 fail exactly on the fixed behaviors, then pass.

## CHANGELOG
The #246 entry's upgrade note said keys whose owner lost roles lose those scopes. That still holds for scopes the owner genuinely lost — but keys no longer need re-minting for scopes the owner *retains*, so the new entry reconciles the two rather than contradicting.

Test-run caveat: the repaired pytest file needs the Flask `app` fixture, which cannot start on Windows (session-dir permission gate). I verified all four fixes through a standalone harness against the real modules (17/17), but their first run under pytest will be CI.


NeySlim (approved):
The code here is correct and mergeable — approving on the merits. Two clarifications for the record.

The 'three gaps' in the title are from your own audit, not the three points from my #246 review — worth noting so this doesn't read as closing that review. Of my points, the wildcard API-key one IS fixed here: the intersection now expands `<cat>:*` to the owner's current scopes in that category instead of dropping a stored `read:*` to zero on demotion, with no way to re-widen (only the owner's effective perms enter the rebind, and the admin `*` case short-circuits earlier). Verified correct.

The two bonus gaps are real and well-fixed: the SSH empty-policy collapse (`[]` → NULL → re-inflated to the full standard set; `if extensions is not None` fixes it, NULL preserved, no migration) and the explicit-ca_id mTLS path (a non-configured CA now requires write:cas, fail-closed when none configured). 62 + 179 regression tests green, /tsa stays public.

Still open from my #246 review, please fold into a follow-up:
1. tsa_require_dedicated_cert is still only settable via direct DB edit — not exposed in GET/PATCH /api/v2/tsa/config or the UI.
2. The EST SSL_CLIENT_VERIFY line in the test module docstring is still there without a test in that module (the fix does now exist via #248, so it's no longer misleading, just uncovered).

NeySlim:

heidrickla:
Closes both of your open #246 review asks, and repairs a claim in this PR's own CHANGELOG that was not true. **2959 passed / 6 skipped / 0 failed** on Linux.

**`tsa_require_dedicated_cert` is no longer DB-edit-only.** You called the toggle "effectively hidden" — it was worse than that: `tsa_service.py` tells the operator to "set tsa_require_dedicated_cert=true" with no supported way to do it. `GET /api/v2/tsa/config` now returns it, `PATCH` accepts it, and the TSA page has a toggle wired like its siblings. Default stays off, so nothing changes for existing deployments. The PATCH test asserts the enforcement read in `TSAService` actually flips, not merely that the field round-trips.

**The Account page now matches the backend gates — and my CHANGELOG claim was false.** I wrote "selecting the configured CA explicitly (as the UI does)". It doesn't: the dropdown listed *every* CA from `casService.getAll()`, so the `write:cas` gate this PR added would 403 an operator the UI had just invited to choose. And per your #246 note, the Generate tab had no permission check at all — any authenticated user saw it, and a viewer got a bare 403.

Both now use the house `usePermission()` pattern (as TSAPage/SCEPPage/ESTPage do): the Generate tab requires `write:user_certificates`, the issuing-CA selector requires `write:cas`, and without it the request simply omits `ca_id` so the backend issues from the configured mTLS CA. Import/Assign stay open to everyone since their routes are bare `@require_auth()`. I deliberately did **not** pin the configured CA's refid client-side — `GET /api/v2/mtls/settings` needs `read:settings`, which the affected principals don't hold, so the UI would have had to fail to find out. CHANGELOG claim struck and replaced with what the code actually does.

**Stale docstring line removed** — the EST `SSL_CLIENT_VERIFY` finding belongs to #248, not #246. You asked for that on the #246 review; my earlier commit here edited that exact docstring and left the line as untouched context, which is worse than not having looked.

Fail-before/pass-after both directions: backend **2 failed → 2 passed** (whole file 27 passed), frontend **4 failed → 6 passed**.

One note on the frontend suite: 511/512, and the single failure is the `AppShell` icon-import timeout — it reproduces with all my changes stashed and passes in a sibling checkout, so it's the local flake that [#250](https://github.com/NeySlim/ultimate-ca-manager/pull/250) raises the timeout for, not a regression here.

PR #254, opened 2026-07-31
fix(signing): close EKU and pathLen gaps left by the issuance hardening

heidrickla:
Follow-up to #243. The most important item is a **regression that PR introduced** — caught before release, but it inverts the control it was written to add.

## 1. A CSR whose EKUs are all refused gets an UNRESTRICTED certificate
When the cap filters out every EKU the CSR requested, `kept` is empty, the `if safe_ekus:` guard adds no ExtendedKeyUsage extension, and the type-default block is skipped because `csr_has_eku` is true. The leaf ships with **no EKU extension at all** — unrestricted for every purpose under RFC 5280 §4.2.1.12.

The incentive is backwards: a CSR requesting *nothing* gets a serverAuth-restricted certificate, while a CSR requesting something *forbidden* gets one restricted by nothing. Asking for more than you're allowed returns more than asking for nothing.

Confirmed as a regression by running the same probe against the pre-#243 file:

| CSR EKUs (`cert_type='server_cert'`) | before #243 | on dev today |
|---|---|---|
| `[serverAuth, codeSigning]` | serverAuth | serverAuth (cap works) |
| `[codeSigning]` only | **codeSigning** (restricted) | **no EKU extension** |
| `[emailProtection]` only | emailProtection | **no EKU extension** |
| `[anyEKU]` only | — | **no EKU extension** |
| none | serverAuth | serverAuth |

Reachable from ACME finalize and EST with no non-default config — the enrollee just changes one line of its CSR. Before #243 that same request produced a *restricted* certificate, so this is strictly a loss.

The existing suite missed it because every case there puts a surviving EKU alongside the forbidden one, so total-drop was never exercised.

Fixed inside `_filter_csr_ekus`: when everything is dropped, fall back to the certificate type's own profile. The helper is **shared with the pre-existing no-EKU default block**, so "requested nothing" and "requested only forbidden things" now resolve identically — that unification is what actually removes the incentive. It also extends profiles to `server`/`host`/`user`/`code_signing`/`email_cert`/`device_cert`, which previously issued EKU-less (i.e. unrestricted) on a no-EKU CSR.

## 2. EST and renewal silently narrow a device's existing EKUs
You raised this on #243 ("auto-renewal drops custom EKUs"). A device whose certificate legitimately carries ipsecIKE or emailProtection loses it on renewal, because these paths pass no `cert_type` and default to the TLS ceiling.

Worth flagging: your suggested fix — propagate the stored `cert_type` — would **not** have worked here. EST rows store `cert_type='server'`, which maps to the same TLS ceiling, so the EKU would still be dropped. So this implements renewal-at-par instead, mirroring SCEP's already-reviewed policy: the ceiling widens by the prior certificate's EKUs, with OCSPSigning/timeStamping/anyEKU/SmartcardLogon excluded from the widening so they can never be resurrected. EST's three endpoints now pass an explicit `device_cert` profile matching SCEP's allow-list.

## 3 & 4. Two pathLen gaps (both low)
The pathLen-0 refusal was bypassed by a CSR carrying **no** BasicConstraints — the shape UCM's own CSR generator emits — because the hardcoded default skipped the clamp. And the sibling route `POST /api/v2/cas` still took `pathLength` straight from user input with no clamp against the parent, leaving the same RFC 5280 §4.2.1.9 violation reachable a different way. Both now route through one shared `capped_path_length()` helper.

## What I deliberately did NOT change
`allow_sensitive_ekus=True` on the two admin sign endpoints. It predates #243, the PR body scoped it out, and `test_admin_sensitive_path_is_uncapped` pins it. The only capability not already reachable through the sanctioned `extra_ekus` route is anyEKU pass-through, and removing CSR-supplied Smartcard Logon there could break a legitimate enterprise smartcard-enrollment flow — that's your product call, not an audit follow-up. I did fix the two module comments that claimed those OIDs are "never honoured from a CSR, whatever the certificate type", which wasn't true of the admin path.

## Tests
Fail-before/pass-after demonstrated by stashing **only the five source files** and re-running: **14 failed, 27 passed**, every failure a new regression test —

```
test_all_ekus_dropped_falls_back_to_type_profile
  AssertionError: certificate has no EKU extension: unrestricted / assert set()
test_renewal_preserves_prior_ipsec_ike
  assert <ObjectIdentifier(1.3.6.1.5.5.7.3.17)> in {serverAuth}
test_pathlen_zero_parent_refuses_csr_without_basic_constraints
  pre-fix signed pathLen 5 under a pathLen-0 parent
```

`git stash pop`, re-run: 50 passed. The other 13 new tests are labeled controls and pass both ways by design.

Two of them are AST tests pinning the EST wiring, in the spirit of the one you liked on #178 — EST endpoint round-trips need the app fixture, which can't start on my Windows host, so those paths are on CI.

## CHANGELOG
I **amended** #243's existing EKU bullet rather than adding a contradicting one — it currently claims the cap works, and that claim needed reconciling. Both are in the same unreleased cycle, so the regression never shipped.


NeySlim (approved):
Approving — and you were right where I was wrong. Propagating cert.cert_type to auto-renewal (my #243 suggestion) wouldn't have worked: the stored type is 'server', which maps to the same TLS cap, so renewal-at-par is the correct fix. More importantly this closes a regression #243 itself introduced: a CSR whose EKUs are all refused produced a cert with *no* EKU extension = unrestricted (worse than over-restricted), reachable via ACME finalize/EST by default. The renew path is safe — widening is limited to prior−allowed with the blocklist (OCSPSigning/timeStamping/anyEKU/SmartcardLogon) never resurrected, and the renewal CSR is still filtered against the widened cap, so an enrollee can't inject an EKU its current cert lacks. The pathLen fixes are real too (the API default CSR shape never triggered #243's pathLen-0 refusal; POST /cas didn't clamp at all under a constrained parent). Shared capped_path_length() applied in both the mixin and before the CA row write. Full suite 2973 passed.

Two follow-ups worth a separate PR:
- SCEP has the same total-drop bug (scep_service.py ~1162: `if safe_ekus:`) — a SCEP CSR requesting only forbidden EKUs still yields an unrestricted cert via auto-approve.
- SCEP renewal can still resurrect OCSPSigning/anyEKU (intentional there, forbidden in sign_csr) — an inter-protocol inconsistency to document or reconcile.

No upgrade note needed since the regression never shipped (same unreleased cycle). For ms-enrollment: this is net-positive — it closes an accidental 'works by having no EKU' path; the feature will need to supply Smartcard Logon explicitly via the template mechanism SCEP #228 already has.

NeySlim:

heidrickla:
Self-review of this PR before you spend time on it found two regressions I had introduced, plus a security control with no test that could kill it. All three fixed here, and this is the first batch verified against a real full run rather than inferred from CI — **2982 passed / 6 skipped / 0 failed** on Linux.

## Two regressions in my own fix
**EST certificates lost the KeyUsage extension entirely.** Passing `cert_type='device_cert'` was fine, but the "Add key usage based on cert type" block only handles `intermediate_ca`/`server_cert` with no `else`, so `device_cert` fell straight through and emitted no KeyUsage at all. Before my PR those certs carried digitalSignature+keyEncipherment. Fail-before was unambiguous: `ExtensionNotFound: No <ObjectIdentifier(oid=2.5.29.15, name=keyUsage)> extension was found`.

**Renewal-at-par was skipped whenever the CSR carried no EKU.** `renewal_of` only reached `_filter_csr_ekus`, which a no-EKU CSR never enters — so the narrowing you flagged on #243 was still happening on exactly the path I claimed to have fixed. Fail-before shows the loss precisely: `assert <clientAuth> in {<serverAuth>}`. The prior certificate's EKUs are now merged in the no-EKU branch too, minus the hard block-list so OCSPSigning/timeStamping/anyEKU/SmartcardLogon can never be resurrected. Deliberately not routed through `_filter_csr_ekus`: an all-blocked prior set on an unprofiled type would hit its empty-kept fallback and hard-fail a renewal.

## The ceiling had no test that could fail
Deleting the entire `device_cert` entry from `_CERT_TYPE_ALLOWED_EKUS` left all my tests green — the control was untested. Now proven by actually deleting it: **3 failed, 14 passed**, exactly the three new guards. One of them, `test_device_ceiling_matches_scep_allow_list`, AST-extracts SCEP's `_ALLOWED_EKU_OIDS` and asserts set equality, so the two hand-copied lists cannot silently diverge — the same pin-to-reality idea as the AST test you liked on #178.

## Your admin-path question, decided deliberately
The uniform default does narrow `code_signing` and `email_cert` on the admin Sign-CSR path. I kept it: an EKU-less leaf is *unrestricted*, so gating the default on `allow_sensitive_ekus` would preserve an unrestricted-certificate generator for precisely the two sensitive types. Admins keep full control through uncapped CSR EKUs and `extra_ekus`. Added the missing `**Upgrade note:**` naming the affected types, and pinned the invariant that every ceiling type's default is non-empty and a subset of its ceiling.

## Correcting the record on SCEP
Your #243 note assumed the EKU cap strips Smart Card Logon from SCEP. It doesn't — SCEP never routes through `TrustStoreService.sign_csr` at all. It builds the certificate itself and applies its own `_ALLOWED_EKU_OIDS`; the `cert_type="server_cert"` you saw there is only the Certificate DB-row label and never reaches `_filter_csr_ekus`. SCEP also already has the profile plumbing you asked for — a profile-bound template's `extended_key_usage` replaces the CSR's and explicitly bypasses that allow-list. EST is the one with no such mechanism, which is worth its own issue rather than smuggling into this PR.

heidrickla:
Thanks for the thorough pass. On your two SCEP follow-ups — I went to port the total-drop fix straight across and stopped, because I don't think it's the same bug, and the difference changes both the priority and the shape of the fix.

**Why it isn't the same severity.** In `sign_csr` the damage came from the *asymmetry*: a no-EKU CSR got the type's default profile, so a CSR requesting only forbidden EKUs ended up **less** restricted than one requesting nothing. Asking for more than you're allowed returned more. SCEP has no such default — `ExtendedKeyUsage` appears in only three places in `scep_service.py` (the allow-list, the template path, and that filtered copy), and nothing supplies a fallback. So a SCEP client can already obtain an EKU-less certificate by simply omitting EKUs from its CSR, which is what most devices send anyway. The total-drop path grants an enrollee nothing it couldn't already get trivially.

That makes it a wart rather than an escalation: the real issue is that **SCEP's default is unrestricted, full stop** — total-drop is just one way to reach a state that's already the norm. Worth fixing, but I wouldn't rank it alongside the `sign_csr` one, and I didn't want you prioritizing it on the assumption that they're twins.

**Which is why I'd rather ask than guess.** Fixing it properly means deciding what SCEP's default EKU *should* be, and that's your call:

1. **Fall back to a profile** (like `sign_csr` now does). Most consistent across protocols, but it changes behavior for every no-EKU CSR too — i.e. most of a device fleet — or, if applied only to the total-drop case, it creates the same asymmetry in reverse.
2. **Refuse the enrollment** when every requested EKU is forbidden. Semantically cleanest: the client asked for something it may not have, so it gets an error rather than a silently different certificate. But SCEP clients handle failure badly, and it turns a working (if over-permissive) enrollment into a broken one.
3. **Default only in the total-drop case.** Smallest blast radius, slightly odd semantics.

My instinct is (2) for correctness with (3) as the pragmatic choice, but you know the device fleets and the ms-enrollment work far better than I do — and this is exactly the kind of change that's invisible until an MDM stops enrolling.

**Happy to do the work** once you've picked, including the renewal/blocklist inconsistency (SCEP deliberately allows resurrecting OCSPSigning on renewal, `sign_csr` forbids it — worth reconciling or documenting as intentional). Tell me which semantics you want and I'll open a separate PR with the 43 existing SCEP tests kept green plus fail-before/pass-after coverage. Equally happy to leave it with you if you'd rather own it alongside ms-enrollment.

Taking the `169.254.170.2` gap now — separate PR shortly, since that one has real impact and no design question.

PR #255, opened 2026-07-31
fix(security): close the confirmed follow-ups from the #248 audit review

heidrickla:
Follow-up to #248, including both items you flagged in your review. One of these turned out to be considerably worse than the audit note suggested.

## 1. Dev compose: the Werkzeug ports really are on 0.0.0.0 (your ask)
You said binding 8443/8080 to loopback "would actually close it". Confirmed: the base file publishes them with no host_ip and the overlay didn't narrow them, so the banner #248 added described a state that wasn't true.

Worth knowing if you patch this yourself — **a bare `127.0.0.1:` prefix does not work.** Compose merges `ports` by (host_ip, published, protocol), so a prefix-only patch *appends* a second mapping and leaves the 0.0.0.0 bind alive. I reproduced that. It needs `ports: !override` (Compose ≥ 2.24, noted in the file).

All published dev ports now bind loopback: 8443, 8080, debugpy 5678, Mailhog 8025/1025. Banner rewritten to describe what is actually exposed — the unauthenticated traceback page, not just the PIN-gated console. Escape hatch for Docker-in-a-VM: `UCM_DEV_BIND=<iface>` moves the app/Mailhog ports; debugpy stays loopback unconditionally with a pointer to `ssh -L`.

Verified by rendering the merged config: after, all five ports show `host_ip=127.0.0.1`; stashed, 8443/8080 show `host_ip=None`.

## 2. OPNsense UI now agrees with the backend (your ask)
The finding pointed at `pages/settings`; the real and only caller is `OperationsPage.jsx`. Checkbox defaults to checked, and rehydration changed from `config.verify_ssl || false` to `!== false` so an old saved blob can't re-poison the default. Opting out now shows the same amber warning as the SSO and Let's Encrypt fields, and the backend logs the opt-out at WARNING.

**Upgrade note for the CHANGELOG:** connection tests against stock self-signed OPNsense certs now fail until the operator deliberately opts out. That is the point, but it will be noticed.

## 3. init_db.py wasn't just inert — the same bug wipes a database
The audit said init_db's hardening was inert and raised TypeError. The cause is a pre-existing `is_active=True` where the model column is `active` (`models/user.py`). But the identical constructor appears in the **database-reset endpoint**, where it runs *after* `drop_all()`:

```
drop_all() → create_all() → User(..., is_active=True)  → TypeError
                                                       → bare except → "Database reset failed"
```

So a reset leaves a **wiped database with no admin user at all**, reporting only a generic failure. Fixed at both call sites.

While there I aligned init_db with the canonical paths per your divergence note: it now seeds the *configured* `INITIAL_ADMIN_USERNAME`/`EMAIL`/`PASSWORD` via `set_password`, with `force_password_change=True` **and** `totp_exempt=True`. Previously it hardcoded `admin`, so any username override meant it tried to seed a duplicate.

## 4. EST log noise
The fail-closed refusal is unchanged; the WARNING now fires only when a certificate was actually presented, with a debug line on the quiet path. Removes a warning per ordinary Basic-auth request behind a same-host proxy, and serverkeygen's double-warning.

## Tests
Fail-before/pass-after on all three testable items:
- EST (5 tests, bare-Flask so they run locally): stashed → `test_basic_auth_request_without_cert_logs_no_warning` fails on the captured spurious record. Four of the five also pin #248's fail-closed behavior, which shipped untested.
- init_db/reset (6 tests, AST guard across all four bootstrap seeding sites): stashed → both sites flagged for `['is_active']`.
- OPNsense UI (5 vitest tests): stashed → all 5 fail.

Compose has no unit test — a meaningful one needs `docker compose` on the runner, and I'd rather say that than commit a hollow stand-in. The rendered-config check above is the evidence.

**Not run locally:** three backend captor tests in `test_opnsense_import.py` use the shared `app` fixture, which can't start on Windows (`app.py` session-dir gate). They mirror the existing tests exactly, but CI is their first real run. I also corrected the two pre-existing stubs there, which had baked `verify_ssl=False` into the old insecure contract.

## One thing outside the strict scope, flagged for your call
Getting the frontend test to run at all exposed two broken shared mocks in `pageRenderingSetup.jsx`: `opnsenseService` mocked `getConfig`/`sync`, which don't exist on the real service, and `useMobile` lacked `screenWidth` — so **every ResponsiveLayout page smoke test was silently rendering only its mobile menu**. I fixed both; the full suite stays green (11 files, 138 tests). Happy to split that out if you'd rather keep this PR narrow.

## CHANGELOG
Amended #248's bullet where it is now false ("binds the debugger to loopback", "the UI checkbox is unchanged") rather than adding contradicting text.


NeySlim (approved):
All the confirmed #248 follow-ups are handled correctly. The dev-compose fix is complete: the base file did publish 8443/8080 on 0.0.0.0 (so #248's banner was misleading), and `ports: !override` is the right tool — a plain `127.0.0.1:` prefix would have added a second mapping while leaving the 0.0.0.0 bind live. All five published ports are loopback now, with a UCM_DEV_BIND escape hatch for Docker-in-VM. OPNsense UI default is flipped to verify=on with sessionStorage re-hydration hardened and the existing sso.sslWarning i18n key reused (no new strings), backend adds a WARNING on opt-out only — no other network-behavior change. The bonus reset_db fix is a real one (the `active` vs `is_active` TypeError left the DB dropped with no admin). Full backend suite 2963 passed, frontend 138.

One note for the CHANGELOG/upgrade text: `!override` needs Compose ≥ 2.24 — an older Compose fails loudly at parse, which is documented in the file and preferable to a silent 0.0.0.0 bind. The .env.docker.example placeholder item from #248 is still cosmetic and unaddressed — fine to drop or defer.

NeySlim:

heidrickla:
Both CHANGELOG points from your review are handled.

**Compose ≥ 2.24 — promoted to a real upgrade note.** It was already in the entry, but only as a parenthetical, and every sibling breaking change in that section carries an explicit `**Upgrade note:**` marker. A hard YAML parse error that stops the dev stack deserves the same treatment, so it now reads as one and states the failure mode plainly: the stack stops with an error naming `!override`, which is the intended outcome rather than a regression — failing loudly beats silently reverting to an all-interfaces bind. Also noted that production compose files are untouched.

**The `.env.docker.example` placeholder is actually done** — your review landed at 23:02Z and I pushed that fix at 23:27Z, so you were looking at the earlier head. `POSTGRES_PASSWORD` is now left **empty** rather than `REPLACE_WITH_A_STRONG_RANDOM_PASSWORD`, because that placeholder was a perfectly valid PostgreSQL password: `cp .env.docker.example .env && docker-compose up -d` stood up a database whose password is public knowledge. Empty makes the official postgres image refuse to initialize ("Database is uninitialized and superuser password is not specified"), so an unedited copy fails at first startup instead of running. That is the difference between your "fails predictably" and actually failing loudly — worth doing rather than deferring, since it costs one line.

The same push also carried the reset_db seeder the rest of the way. You spotted the `active`/`is_active` TypeError; the same handler was also hard-coding `admin`/`admin@localhost` instead of the configured identity, omitting `totp_exempt`, and bypassing the pinned password hasher — so under global 2FA enforcement a UI database reset recreated an admin nobody could log in as. All three fixed, and the guard test strengthened, since it previously only checked kwarg *names* and passed green on all of them.

Full backend suite on the current head: **2970 passed, 6 skipped, 0 failed.**

PR #256, opened 2026-07-31
fix(acme): refuse metadata targets and malformed DNS identifiers in challenges

heidrickla:
Follow-up to #244, leading with the item you asked for in your review.

## 1. Cloud metadata is now refused unconditionally (your follow-up)
You wrote: with `allow_private_ips` on (the default), an order for `169.254.169.254` still triggers a challenge GET to the metadata endpoint, whereas the project's convention elsewhere blocks metadata even when private ranges are allowed — and that an unconditional check in both validators would be a good follow-up.

Done, using the existing house helper `validate_url_not_cloud_metadata`, called **before** the private-IP guard in both validators with `allow_loopback=self._acme_allow_private_ips()`.

That parameterisation is the whole trick for not breaking on-prem: RFC1918 addresses, `.lan` names and loopback behave exactly as today when the setting allows them. Only metadata endpoints are refused unconditionally — 169.254.169.254, 100.100.100.200, `fd00:ec2::254`, `metadata.google.internal`, and an IP dressed as a name (`169.254.169.254.`). TLS-ALPN-01 has no URL of its own, so the authority is synthesized through the existing `format_ip_for_url`.

## 2. Malformed `dns` identifiers (the confirmed finding)
A `dns` identifier carrying a port or userinfo evaded the guard. Rather than patch the validators, I fixed it at the door: `validate_acme_identifier` now syntax-checks `dns` values (LDH plus underscore, ≤253, tolerating `*.` and one trailing root dot). That covers **newOrder and newAuthz both**, and rejects the whole class — `:`, `@`, `/`, `?`, `#`, whitespace — instead of the two sample strings.

Two defence-in-depth changes alongside it: `validate_host_not_private` now **fails closed on `gaierror`** instead of falling through, and strips a trailing root dot before the literal-IP check so `169.254.169.254.` is classified as the link-local literal it is.

## 3. DNS rebinding — deliberately partial, and here's why
Where the guard actually runs (`allow_private_ips=false`) the validated IP is now pinned: HTTP-01 fetches inside the existing `pin_host` contextmanager, TLS-ALPN-01 connects straight to the validated address (SNI was already decoupled).

**I did not pin the default configuration**, and left a comment saying so. There, private answers are legitimate, so pinning would buy only the metadata check — which the unconditional check above already covers — while collapsing every on-prem validation onto a single address with no dual-stack failover. That's a bad trade for the product's primary use case. Happy to revisit if you disagree.

I also amended the stale note on `pin_host` saying "do NOT use for ACME" — no longer true.

## Note on paths
Both cited paths in the audit were stale: the validators are in `services/acme/mixins/challenge.py` (not `challenge_validators.py`) and the guard is `utils/ssrf_protection.py`. Everything below was re-verified against the real code by execution.

## Tests
New `backend/tests/test_acme_dns_identifier_ssrf.py` — 43 tests driving the mixin through a small harness rather than the `app` fixture, so they run locally **and** in CI, with every DNS answer stubbed so a hijacking resolver can't decide the outcome.

Fail-before/pass-after by stashing only the source files: **30 failed, 13 passed**, with assertions naming the exact egress —

```
the validator reached out to 'http://10.0.0.5:8080/.well-known/acme-challenge/...'
the validator reached out to 'http://evil@10.0.0.5/.well-known/acme-challenge/...'
the validator reached out to 'http://169.254.169.254/.well-known/acme-challenge/...'   <- with allow_private_ips=True, your case
```

After: **43 passed**. The 13 that pass both ways are controls — LAN, loopback and private targets must still be fetched.

I also tightened the two existing blocked-path tests in `test_acme_ip_identifier_ssrf.py` that a reviewer flagged as passing for the wrong reason; they now assert `challenge.status == 'invalid'` and `rejectedIdentifier`, not merely a falsy return.

**Caveat:** those two edited tests use the `app` fixture and could not run on my Windows host — they're proven only in CI. The full backend suite shows 4 failures here, identical with my changes stashed (Windows-only: a cp1252 decode, the PG migration runner, two scheduler `OSError`s), so no regression.


NeySlim (approved):
This closes the residual hole I flagged on #244, exactly as suggested and without breaking the LAN. Both challenge validators now call validate_url_not_cloud_metadata before and independently of the private-address guard, with allow_loopback tied to the private-IPs setting — so RFC1918, .lan and loopback still pass under the default (confirmed by a dedicated negative-control test that fails if a legitimate private target is blocked), and only metadata/link-local literals are refused unconditionally. Malformed-DNS rejection is LDH+underscore per label with no false positives on normal .lan/.local names. Coverage spans IMDS v4/v6, the cloud metadata hostnames, and hostnames resolving to metadata at check time. 55/55 tests green.

Two follow-up notes:
- 169.254.170.2 (AWS ECS task-credentials endpoint) isn't in _CLOUD_METADATA_IPS — same credential-theft impact as IMDS; pre-existing helper gap, easy add.
- Under the default, the fetch isn't pinned, so a short-TTL rebind (public at check, metadata at fetch) stays theoretically open — you noted this yourself and the trade-off is reasonable for a LAN product.

NeySlim:

heidrickla:
Self-review found three defects in my own PR, one of which made it reject legitimate hostnames. All fixed. **3040 passed / 6 skipped / 0 failed** on Linux.

**1. I decided policy from an exception message.** `except ValueError` caught both my deliberate policy raise and `ipaddress.ip_address()`'s *parse* failure — whose message embeds the hostname verbatim. So any ordinary host containing "metadata", "loopback", "unspecified", "private" or "reserved" as a substring was refused: `metadata-db.corp.example.com`, `loopback0.rtr1.corp.example.com`, `private-ca.corp.example.com`. Parsing and policy are now separated at **all three** sites (`validate_host_not_private`, `validate_url_not_cloud_metadata`, `_resolve_and_validate`) — the verdict comes from the parsed address or not at all. Every genuine deny-list target is still refused, pinned by a test.

**2. I left the fail-open I had just disproved, in the guard I made load-bearing.** This PR changed `validate_host_not_private` to fail closed on `gaierror`, on the reasoning that `169.254.169.254:80` is unresolvable as a hostname but reachable once `requests` re-parses it. But `validate_url_not_cloud_metadata` still had `except socket.gaierror: pass` — and the same PR made *that* the unconditional check while leaving the other gated behind `allow_private_ips`, which defaults to **true**. In the default configuration the only guard running still carried, verbatim, the comment this PR proved wrong. Now fails closed, with E2E tests driving both challenge validators in the default config.

**3. Pinning discarded N-1 already-validated addresses.** `validate_host_not_private` raises if *any* resolved address is private, so when it returns, every address is public-safe — keeping only the first threw away dual-stack failover for no security benefit. The whole validated set is now pinned and failed over within, never falling back to the OS resolver while pinned.

**Please look closely at one thing.** Failing closed on NXDOMAIN exposed 25 existing tests that only passed because made-up hostnames slipped through the fail-open — they were resolver-dependent all along, passing or failing on whether your DNS hijacks NXDOMAIN. Rather than weaken the guard I added an autouse fixture in `backend/tests/conftest.py` that replaces **only `ssrf_protection`'s** view of `socket`: real answers win, and only a resolution *failure* falls back to a fixed TEST-NET-3 address. Everything else keeps the real resolver. SSRF tests that need a specific answer or a genuine failure monkeypatch it themselves and fully override the fallback.

That is a suite-wide test-infra change inside a security PR and I would rather you weigh it than discover it. My first attempt patched `socket` globally and made the suite hang on unroutable connects (318s, 1 failure); the scoped proxy restored normal behavior (160s, 0 failures). If you would prefer those 25 tests pinned individually instead, say so and I will do that.

Fail-before/pass-after: **39 failed → 117 passed** across the four SSRF test files. CHANGELOG corrected — the earlier "exactly as before" claim was wrong — with an explicit upgrade note for the fail-closed change.

heidrickla:
Added `169.254.170.2` — good catch, and you were right that it's the same impact class as IMDS. **3043 passed / 6 skipped / 0 failed.**

Why it was missing is worth recording, because it explains the shape of the fix: it is not the *instance* metadata service, so it didn't look like the thing being listed. A container reaches it by appending `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`, and it returns live task-role IAM credentials exactly as IMDS does. Blocked unconditionally with the rest — a link-local address is never a legitimate ACME, webhook or SSO target whatever `allow_private_ips` says.

Fail-before/pass-after: reverting only `ssrf_protection.py` gives **2 failed, 91 passed**, both the plain and the IPv4-mapped IPv6 form (`http://[::ffff:169.254.170.2]/`), then **93 passed** restored.

**One honest note about my own test.** I added a second test that derives its cases from `_CLOUD_METADATA_IPS` rather than restating them, and in the commit message I implied that guards against this recurring. Running the fail-before showed it does not: remove the entry from the source and the derived test simply iterates one fewer address and passes. It guards a different property — that every listed address is actually *enforced* under the permissive setting — which is worth having, but it is the explicit parametrised case that catches an address being *absent*. Both are needed and they cover different failure modes. Flagging it because I nearly shipped a claim the evidence doesn't support, which is the same thing I've been auditing for all day.

On the other note you left here — the short-TTL rebind under the default configuration — I agree the trade-off is reasonable and haven't touched it. Pinning in the default config would buy only the metadata check, which the unconditional guard now covers anyway, while collapsing every on-prem validation onto a single address with no dual-stack failover.

Your two SCEP follow-ups are on #254; short version is I don't think the total-drop one is the same bug and I've asked which semantics you want before writing it.

PR #257, opened 2026-08-01
fix(cert,scep): read subject_cn from the subject, not its RFC 4514 string

heidrickla:
Your #251 follow-up. Turned out to be two copies rather than one, so it seemed worth doing properly instead of someday.

## The bug
`subject_cn` was derived by splitting `subject.rfc4514_string()` on `,` and keeping the first `CN=` token. RFC 4514 §2.1 emits RDNs in **reverse** order, so that token is the **last** commonName:

```
subject:            CN=web.example.com, CN=admin.example.com
rfc4514_string():   CN=admin.example.com,CN=web.example.com
first CN= token:    admin.example.com     <- recorded and displayed
ACME finalize:      web.example.com       <- validated
```

The name **checked** and the name **recorded** disagreed exactly when a subject carried more than one CN — the case an attacker controls. The split was independently wrong too: RFC 4514 escapes a comma inside a value as `\,`, so `O=Example, Inc.` could truncate the name.

You're right that #251 closed the ACME route by refusing multi-CN subjects at finalize. But the derivation still fed the other signing paths, and it had drifted into **two** copies — `services/cert/mixins/csr.py` and `services/scep/scep_service.py`. So the defect outlived its most dangerous caller in two places at once.

## The fix
One helper, `utils.dn_parse.subject_common_name()`, called by both. I put it next to the existing string-based `get_dn_attribute()` with a note on which to use when: the string form unescapes correctly but **cannot recover ordering**, so the object form is the only correct choice whenever an x509 Name is in hand.

SCEP's value feeds the `Certificate` row's `subject_cn` only — issuance is untouched. No migration: existing rows keep their recorded name, and re-issuing corrects it. I'd rather not rewrite historical rows silently; say the word if you'd prefer a migration.

## Tests
Fail-before/pass-after with the helper present and only the call sites reverted: **1 failed, 8 passed** → **9 passed**. Full backend suite **2963 passed / 6 skipped / 0 failed** on Linux.

**One thing I'd rather tell you than have you find.** My first draft of this test module defined its own copy of the derivation and asserted against *that* — so it would have passed no matter what the production code did. Exactly the tautology this whole batch has been auditing for, in the test I wrote to close it. The committed version drives the real helper, and there's a guard test that fails if either recorder grows a third copy of the split.

## Not done
Eight other call sites split a subject **string** to find a CN (`api/v2/search.py`, `sso/mapping_tests.py`, `truststore.py`, `user_certificates.py`, `smart_import/{importer,validator}.py`, `migrate_v1_to_v2.py`). Those receive a string with no x509 object in reach, so they are a different problem: they inherit whatever the producers recorded, and fixing the producers is what actually matters. They do share the `\,` escaping weakness, which is worth a separate pass if you want it — I didn't want to bundle a nine-file refactor into a two-line correctness fix.


NeySlim (approved):
Verified against dev: rfc4514_string() does emit RDNs in reverse order, so the old split recorded the last commonName while ACME finalize validates get_attributes_for_oid(COMMON_NAME)[0] — the first — and the two disagreed exactly on the attacker-controlled multi-CN case. The split also ignored the `\,` value-escape. Both recorders (services/cert/mixins/csr.py and services/scep/scep_service.py) now route through the single subject_common_name() helper, which is the correct object-based derivation.

The test module is the right kind: it drives the real helper, pins the premise (that cryptography reverses), compares against the same cryptography API ACME uses rather than a re-implementation, and adds a guard that fails if a third copy of the split reappears. Appreciate you calling out and fixing the tautological first draft — that's the exact failure mode this batch has been about.

Full backend suite green on the branch (2962 passed), no overlap with the SCEP EKU fix that just landed. No migration is the right call — silently rewriting historical rows would be worse; re-issuance corrects a row, and the note says so. This closes the subject_cn derivation point I raised on #251. The eight string-splitting call sites you list are correctly out of scope.

Issue #260, opened 2026-08-01
[Bug][Critical][Security Issue] ACME Proxy Endpoints have missing ownership-check


gb-123-git:
The proxy's authz, challenge, get_order, and cert endpoints verify the JWS signature but never check that the requesting account owns the resource:
```
# proxy authz — no account ownership check
def authz(authz_id, slug=None):
    is_valid, payload, _, err = verify_proxy_jws()  # ✅ JWS verified
    # ❌ No check: does this account own this authorization?

# proxy challenge — no account ownership check  
def challenge(chall_id, slug=None):
    is_valid, payload, _, err = verify_proxy_jws()  # ✅ JWS verified
    # ❌ No check: does this account own this challenge?

# proxy get_order — no ownership check
def get_order(order_id, slug=None):
    is_valid, payload, _, err = verify_proxy_jws()  # ✅ JWS verified
    # ❌ No check: does this account own this order?

# proxy cert download — no ownership check
def cert(cert_id, slug=None):
    is_valid, payload, _, err = verify_proxy_jws()  # ✅ JWS verified
    # ❌ No check: does this account own this certificate?
```
The Native ACME Server checks:
```
# native authz
if auth_account != request_account_id:
    return acme_error('unauthorized', 'Authorization does not belong to this account', 403)

# native order
if order.account_id != request_account_id:
    return acme_error('unauthorized', 'Order does not belong to this account', 403)
```

**Impact: Any registered account on the proxy can read any other account's authorization status, challenge details, order status, and download their certificates by guessing/enumerating ID**

Example:
```
Bob → POST /acme/proxy/authz/<alice_authz_id>
     → JWS signed with Bob's key, kid points to Bob's account
     → ✅ JWS verification passes (Bob's key is valid)
     → ✅ Returns Alice's authorization details (no ownership check)

Bob → POST /acme/proxy/cert/<alice_cert_id>
     → ✅ Returns Alice's certificate (no ownership check)

Bob → POST /acme/proxy/order/<alice_order_id>
     → ✅ Returns Alice's order status (no ownership check)
```

NeySlim:
Fixed in 91414d525 on `dev`.

All order-scoped proxy endpoints (`authz`, `challenge`, `order`, `cert` — and `finalize`, which already had the check) now enforce the owner binding recorded at new-order (account id and/or client JWK thumbprint) against the verified requester identity:

[...]

PR #263, opened 2026-08-03
Security Fixes

gb-123-git:
#### CHANGE LOG:

**New Features :**

**A. UX :**
**Issue SSH Certificates - Autosync Certificate Type when SSH CA changes -**
_The Certificate Type automatically changes to the correct type based on the CA._

**B. Security:**
The MR contains several critical security fixes [Closes Issue #260] :

| S.No. | Vulnerability | File-Name |
| :---: | :--- | :--- |
| 1. | Path Traversal in `download_update` allowed file creation in arbitrary paths. | _backend/services/updates.py_ |
| 2. | SQL Injection possibility in `_reset_pg_sequences` | _backend/services/database_admin/helpers.py_ |
| 3. | Missing Path Verification in Backup Download/Delete | _backend/api/v2/system/backup.py_ |
| 4. | Missing Ownership Checks on ACME Proxy Resources | _backend/api/acme/acme_proxy_api.py & backend/services/acme/acme_proxy_service.py_ |
| 5. | Missing Account Status Check in `new_order` (when EAB off) | _backend/api/acme/acme_proxy_api.py_ |
| 6. | DNS Automation Triggered Without Ownership Check in `get_authz` | _backend/services/acme/acme_proxy_service.py_ |
| 7. | Loose Challenge-to-Order Matching with Dangerous Fallback | _backend/services/acme/acme_proxy_service.py_ |
| 8. | IDOR in `revoke_user_certificate` - Missing Ownership Check | _backend/api/v2/user_certificates.py_ |
| 9. | IDOR in `delete_user_certificate` - Missing Ownership Check | _backend/api/v2/user_certificates.py_ |
| 10. | WebSocket Authentication Bypass - Unauthenticated Access to Events | _backend/websocket/events.py_ |
| 11. | WebSocket Missing Room Authorization - Any User Can Subscribe to Any Room | _backend/websocket/events.py_ |
| 12. | SSO Provider Secrets Exposure - Missing Admin Check on `include_secrets` | _backend/api/v2/sso/providers.py_ |
| 13. | CA Export Password in URL Query String - Sensitive Data in Access Logs | _backend/api/v2/cas/export.py_ |
| 14. | LDAP Bind Password stored in plaintext | _backend/api/v2/settings/ldap.py_ |
| 15. | Security-Sensitive Settings Modifiable without Admin Permission | _backend/api/v2/settings/general.py_ |

**Bug Fixes:**
1. Fixed Download Certificates (SSH CA Section)
2. Added missing openssh-client dependency to Dockerfile (required for KRL Generation)

____
### Notes:

**Known limitation:**  _fail-open on legacy orders without ownership binding_

`_verify_order_ownership` currently grants access when a proxy order exists but has no account_id or client_jwk_thumbprint — i.e., legacy rows created before ownership tracking was introduced. This is an intentional compatibility bridge to avoid breaking existing orders during upgrade.

____
_The work done in this PR has been sponsored by PMGA Tech LLP._


gb-123-git:
@NeySlim This is a big security fix, with lots of files changed. I will probably update this next week.

NeySlim (changes requested):
Thanks for this work — I verified the reported issues against the current `dev` branch and most are real (WebSocket anonymous subscriptions, the IDOR on user-certificate revoke/delete, the plaintext private key in the approval issuance path, SSO `include_secrets` exposure, CA export password in the query string, LDAP bind password). The fixes for those are correct in approach. There is however one blocking bug in the new code, plus a few things to clean up before this can move forward.

## Blocking — `respond_challenge` ownership lookup never matches, and breaks the endpoint

In `services/acme/acme_proxy_service.py::respond_challenge`, the new ownership lookup derives the authz URL with `chall_url.rsplit('/', 1)[0]` and matches it against `upstream_authz_urls` with `.contains(...)`. On Let's Encrypt (and on this PR's own test stub), challenge and authorization URLs live in disjoint namespaces — e.g. authz `…/acme/authz-v3/12345` vs challenge `…/acme/chall-v3/12345/AbCdEf` — so the derived URL is never a substring of the stored authz URL and `order` is always `None`. Two consequences:

1. The ownership check never fires: the challenge vector of #260 remains open (the request is forwarded upstream regardless of who asks).
2. Worse: the endpoint now returns `500 serverInternal` for the **legitimate owner** too, because with `order = None` the code falls through to `find_provider_for_domain("unknown")`. This regresses `respond_challenge` entirely — the previous `_find_order_for_challenge` call matched by CA host and worked.

The PR's own test proves it: `test_challenge_cross_account_denied` currently fails (500 instead of 403) — CI is red.

Suggested fix: match on the stable part shared by both URLs (the authz/serial id segment), or restore a dedicated `_find_order_for_challenge` that strips the two known suffix namespaces, and add an **owner happy-path test** for `/challenge/:id` so a regression like this can't slip through again (currently only the cross-account denial is tested).

## Must-fix before merge

- **Remove the "Patch sponsored by PMGA Tech LLP" comments** from the source (they appear in `acme_proxy_api.py`, `acme_proxy_service.py`, `updates.py`, `helpers.py`, `user_certificates.py`, …). This project doesn't carry sponsorship notices in code comments.
- The `_verify_order_ownership` fail-open for orders with no ownership binding (legacy rows) is acceptable as a compatibility bridge, but please note it explicitly in the PR description so it can be revisited (a hard cut-over with a logged deprecation window would be better long-term).

## Minor

- A large part of the diff is trailing-whitespace churn in `acme_proxy_service.py` — fine to keep, but it makes the security-relevant changes harder to review; next time splitting cosmetic-only changes into a separate commit helps a lot on a sensitive diff like this.

Happy to re-review quickly once the challenge lookup is fixed — the rest of the ownership logic (fail-closed when a requester identity is missing on a bound order, ordering of the checks) looks sound, and the new test files (`test_user_cert_idor.py`, `test_sso_secrets.py`, `test_settings_admin.py`) are good coverage.

gb-123-git:
@NeySlim - Appreciate your suggestions. This PR is still a work in progress, that's why I have marked it as 'Draft' for now. I'll be working on it next week.
`services/acme/acme_proxy_service.py` is the one i need to fix and also re-check all the work (since this is a fairly large PR) before I re-open this PR for merge.
I think I need to squash a few more bugs that I am finding while testing the software. I'll update the description once the code is complete for delivery.

As for the sponsor messages, the LLP is sponsoring me for the work. They are not claiming ownership, just want a little 'advertisement'. I'll remove them as per your wish in the code. I understand your reasons.
Where do you think I should put their contribution ?
Or maybe I was wondering if you can reward them in some non-monetary way eg. (maybe grant them an unrestricted license)? Basically if we can get them to sponsor me to work on this on a regular basis, it would be beneficial as I would be able to spend more time on this. Waiting for your thoughts on this.

NeySlim (commented):
Went through the whole diff and ran the full backend suite against this branch locally: 3350 passed, 0 failed. Overall solid work — the websocket fix in particular closes a real hole (any unauthenticated client used to receive every cert/CA broadcast), and on the ACME side resolving the challenge's order through the Link rel="up" header and dropping the "most recent pending order" fallback is the right approach.

A few things before this leaves draft:

**1. Operator UX regression on the settings page**
The admin-only key check is fine on the backend, but the frontend saves whole cards: `SECTION_FIELDS.generalSession` in SettingsPage.jsx sends `session_timeout`, `max_login_attempts` and `lockout_duration` together with `timezone`/`date_format`/`show_time`. So an operator changing just the timezone now gets a 403. Either the frontend needs to filter admin-only keys for non-admins, or the backend should only reject admin keys whose value actually changes.

**2. WebSocket needs a real-browser check**
The old code had that "session not shared with websocket" comment — if any deployment still ends up in that situation, the new code silently refuses every frontend connection and realtime updates just stop. `withCredentials: true` is there on the client so it should work, but please confirm you've seen the connected indicator + live events in an actual browser against this branch.

**3. Description says 16 fixes, the diff has 15**
Item 14 (approval-issued private keys, policies.py) isn't in the diff — dev already encrypts those keys, so your merge from dev picked it up. Worth updating the description.

**4. Sponsor mentions in code comments**
There are half a dozen "sponsored by PMGA Tech LLP" comments in the source (one "sponsired" typo). Happy to keep the credit in the PR description, but please drop them from the code itself.

Minor:
- `get_provider` checks `role != 'admin'` while the websocket code resolves privileges from permissions — would be nicer to use `has_permission` here too so custom roles behave consistently.
- The `startswith(authz_url)` fallback in `_find_order_for_challenge` will never match on LE/Pebble (chall/authz namespaces are disjoint, as the docstring itself notes). Fine as a dead-end fallback since the Link-header lookup covers it, but it could just be dropped.
- The `if local_order is None: return` in `_verify_order_ownership` is unreachable (every call site raises before it) and reads as fail-open — a raise would be safer if it ever becomes reachable.
- The ldap bind_password encryption: nothing currently reads `ldap_bind_password` back from SystemConfig (the GET on that endpoint is a static stub), so this one is mostly cosmetic.
- Quite a lot of whitespace-only churn in acme_proxy_service.py inflating the diff.

gb-123-git:
@NeySlim 
I haven't finished my work as yet.
> 1. Operator UX regression on the settings page

Will try and work on this once I finish the other files.

>2. WebSocket needs a real-browser check

I am still working on the files, can you please do the honors and check it for me ?

>3. Description says 16 fixes, the diff has 15

I had checked the code against the main branch initially. But since the dev branch already has it (due to a PR) I have removed it from the Merge description,

>4.  Sponsor mentions in code comments

I am removing them as I am reviewing the files. Thanks for adding it in the PR description.

>Minor Issues:

Will try and get on it once I fix the major ones.

NeySlim:
WebSocket real-browser check done (item 2 from my last review): **works**

Spun up this branch exactly as production runs it (gunicorn + the gevent websocket worker, not a dev server), drove a real Chromium session: logged in, confirmed the socket upgrades to WebSocket transport with the session cookie, the server answers the `connected` event, and a CA created from a *separate* client session arrives live in the browser as a `ca.created` frame — no polling, no page reload. Anonymous probe (no session) gets the Engine.IO handshake only and is never admitted to the namespace. So session auth, scoped rooms and cross-client broadcast are all behaving as intended.

Also re-reviewed the new commits since my last pass (up to `c1e05ca`):
- `general.py` — stripping admin-only keys for non-admin submitters while still 403-ing pure-admin payloads, plus the frontend locking of admin fields with the lock hint: exactly the resolution I had in mind, thanks. Verified an operator can still save the timezone card and that the stripped attempt is logged.
- `providers.py` `has_permission('admin:system')`, dropping `_find_order_for_challenge` (fail-closed on missing Link header is the right call), and `_verify_order_ownership` raising instead of returning: all good.
- Full backend suite re-run against the branch: 3350 passed, 7 skipped; frontend: 546 passed.

Two leftovers:
- The PMGA Tech LLP sponsor comments are still in `api/v2/user_certificates.py` (2×) and `services/updates.py` — those are the last ones I can find. Once they're out I have nothing left on my side.
- `test_settings_admin.py` churn: current state passes locally, but please make sure CI agrees.

Once the sponsor comments are dropped, good to merge from my point of view.

gb-123-git:
@NeySlim : Removed the PMGA text ( somehow I missed that before my previous submission )
This PR should be good for now. 😅

NeySlim (approved):
All items from both reviews are now addressed: settings UX split works both ways (backend strip + frontend locking), `has_permission` consistency, dead-code removal fail-closed, WebSocket real-browser check passed (gunicorn + gevent worker, session auth, live cross-client broadcast, anonymous rejection), sponsor comments removed. Backend 3350 passed / frontend 546 passed locally; CI green on `5c877a361`. Approving — thanks for the thorough work.

gb-123-git:
@NeySlim :  You're Welcome ! I like your project a lot ! Any ETA on merging this to main and releasing a version ? Since this MR has critical security fixes, imho this should be released on a priority basis.

NeySlim:
Yeah it's merged already, waiting for another PR from @Hemsby to finish and cut a rc.

PR #272, opened 2026-08-11
fix: populate CN from IP SAN when CSR subject is empty

### LINKED ISSUE #271: possible acme issue
currently, I’m having an issue where when I’m trying to use a certificate generated with Certbot it is not showing up as valid in Webmin i’m not exactly sure what the issue is. It is possible this issue stems from Webmin itself but I think that’s unlikely because manually generated certificates work perfectly fine, and I no idea what the issue is here

Hemsby:
## Summary
- `sign_csr`'s empty-subject fallback filled the CN from the CSR's first SAN, but only checked `x509.DNSName` entries. certbot 2.x sends subject-less CSRs by default, so a CSR whose only SAN is an IP address (e.g. `IP Address:10.10.5.54`, as reported in #271) sailed through with a genuinely empty Subject.
- Extends the existing fallback in `csr_operations_mixin.py` to also handle `x509.IPAddress`, stringifying the address into the CN, mirroring the DNSName behavior. Covers every issuance route that shares this code path (ACME, admin sign-CSR, EST/SCEP, WSTEP).

Fixes #271

## Test plan
- [x] New regression tests in `backend/tests/test_issue_271_ip_san_cn_fallback.py`: empty-subject + IP-only SAN populates CN from the IP; empty-subject + DNS SAN still populates CN from the DNS name (no regression on the existing behavior); first-SAN-wins ordering when both an IP and a DNS SAN are present.
- [x] `pytest tests/test_issue_271_ip_san_cn_fallback.py` and `tests/test_csr_eku_profile_cap.py` pass locally.

### DIFF
--- a/backend/services/trust_store/csr_operations_mixin.py
+++ b/backend/services/trust_store/csr_operations_mixin.py
@@ -385,7 +385,7 @@ class CSROperationsMixin:
             ca_cert, csr.subject, csr_sans, renewal_of=renewal_of
         )
 
-        # If CSR has empty subject, populate CN from first SAN DNS name
+        # If CSR has empty subject, populate CN from first SAN DNS name or IP address
         subject = csr.subject
         if not list(subject):
             try:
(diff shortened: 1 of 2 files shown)


NeySlim (approved):
Minimal and correct — extends the existing unconditional CN-from-SAN fallback to IPAddress, with the subject-order case covered by tests. Closes #271's UCM-side hardening.

PR #273, opened 2026-08-11
fix(wstep): check gssapi, not just spnego, for Kerberos availability (GH #229)

Hemsby:
## Summary
- `is_library_available()` only checked `import spnego`, which succeeds even without the actual GSSAPI backend since pyspnego is pulled in transitively by `requests-ntlm` (already required for the WinRM admin channel). That let XCEP advertise the Kerberos CAURI and WSTEP attempt real negotiation instead of the clean 503 the code comments promised, crashing mid-handshake with a confusing `GSSAPIProxy requires the Python gssapi library` error instead.
- Reported by @stefanelul2000 on #229, who also hit the same failure in the CI run for PR #230's merge.
- Now checks `import gssapi` too, with logging at both 503 call sites so "not installed" and "not configured" stay distinguishable, matching `is_configured()`'s own docstring.
- Added a `skipif` guard (matching the existing `pkilint` optional-dependency pattern) to the 4 `test_kerberos_keytab_validation.py` tests that need a real GSSAPI install to pass — these were failing unconditionally in CI since `gssapi` is intentionally kept out of the default requirement set.

## Test plan
- [x] `test_negotiate_auth.py` — added regression test for `is_library_available()` requiring `gssapi`, not just `spnego`
- [x] `test_kerberos_keytab_validation.py` — 4 tests now skip cleanly without `gssapi` installed (locally: 8 passed, 4 skipped)
- [x] Ran full backend suite for regressions — no new failures beyond pre-existing, unrelated, environment-specific ones (Windows-only test-fixture permissions, Postgres-dependent tests)


NeySlim (approved):
Verified: the gssapi-first check is the right fix (base spnego comes in transitively via requests-ntlm, so the old check advertised a broken Kerberos binding), the 503 split keeps 'not installed' vs 'not configured' distinguishable, and the regression test covers it. I retargeted the base to dev and resolved the overlap with the equivalent skipif already merged; CI is green on the result.

PR #274, opened 2026-08-12
feat(wstep): add per-template pinned subject fields (hybrid subject template)

Hemsby:
## Summary

Templates can now pin fixed `O`/`OU`/`C`/`ST`/`L` subject fields that override whatever a client's CSR or AD-derivation supplies on WSTEP issuance, while CN/SAN stay dynamic. This is the "hybrid subject template" request from a follow-up comment on #229:

> a template should be able to enforce fixed organizational fields (C, ST, L, O, OU) from the template itself regardless of what the client or AD-derived subject provides, while still letting CN/SAN come dynamically from AD or the client's CSR [...] a client-controlled or partially-AD-derived request could otherwise omit or misrepresent fixed org identity fields.

## Why this is scoped the way it is

- `dn_template` looks like it should already do this, but it's inert for enforcement — nothing on the backend ever reads it to build a certificate's subject. The only place it's used is the frontend, to prefill an *editable* manual-issuance form. So this needed a new mechanism, not an extension of a working one.
- Deliberately WSTEP-only, not ACME/EST/SCEP/admin issuance. EST has no template concept at all, and forcing this onto the other protocols would mean silently rewriting a requester's own submitted CSR on paths that have never done that — a different, riskier feature than what was asked.
- This isn't ADCS interop parity. Stock ADCS has no per-template "pin these RDNs regardless of source" toggle — real deployments wanting this write a custom CA policy module. Calling that out so it's not read as mirroring existing ADCS behavior.
- Overrides client-supplied values on every WSTEP issuance (Kerberos-bound, naked or manually-typed CSR, and UsernamePassword-bound), not just a naked-CSR fallback fill — matching the reporter's own wording rather than a narrower version that wouldn't actually stop a client misrepresenting its org fields.
- Enforcement lives entirely in `wstep_service.py`, merging into the same `override_subject` mechanism the existing AD-derivation already uses — no changes needed to the shared signing path, so it inherits the existing NameConstraints-ordering guarantee for free.
- Per-field, per-template opt-in — a template with nothing pinned behaves exactly as before. Conflict handling on the CA-wide-fallback case (ambiguous naked-CSR template matching, same gap the Enroll ACL work hit): if candidates disagree on a field's value, that field is skipped and a warning logged, rather than silently picking a winner.

## Testing

- 32 new tests (subject-merge unit tests, template resolution/conflict handling, end-to-end issuance, template CRUD validation). Full backend suite: 3513 passed, 6 skipped, 0 failed.
- Real lab verification against dc1/win11: retriggered actual GPO machine autoenrollment (naked CSR) — issued cert carried both the AD-derived CN and the pinned fields together. Then built a real CSR via `certreq.exe` with a self-declared `O=Evil Corp` and submitted it through the WSTEP endpoint — the issued cert came back with the pinned `O=Acme Corp` instead, CN untouched. That's the actual security property this PR exists for, confirmed against a real Windows-generated request, not just a unit test.

## Known limitations

WSTEP-only by design (see above) — does not apply to ACME/EST/SCEP or admin manual issuance.


NeySlim (approved):
Solid design: pins constrained to O/OU/C/ST/L (CN/SAN untouched), the naked-CSR fallback correctly unions across all CA-resolvable templates, and conflicting values are skipped with a warning rather than picking a winner. Validation goes through validate_dn_field, the migration is dual-backend/idempotent, i18n is in sync across all 9 locales. Backend suite green on my end too (3512 passed).

PR #275, opened 2026-08-12
feat(wstep): embed AD SID security extension for KB5014754 strong mapping

Hemsby:
## Summary

WSTEP-issued certificates now carry the SID security extension (`szOID_NTDS_CA_SECURITY_EXT`, `1.3.6.1.4.1.311.25.2`) whenever the requester authenticated via Kerberos and the AD Connector can resolve their SID — this is what Windows Server's KB5014754 "strong certificate mapping" (mandatory Full Enforcement since Feb 2025) checks for. Answers a feasibility question from `stefanelul2000` on #229.

Unlike the pinned-subject-fields PR (#274), **this one is real ADCS parity, not a new capability** — Enterprise CAs have auto-embedded this exact extension since a 2021-era hotfix, specifically so admins don't have to hand-populate `altSecurityIdentities`.

## Design

- **Unconditional, not a per-template opt-in** — matches real ADCS's own automatic behavior. Every other AD Connector feature in this codebase (`ad_derived_subject`, `allowed_ad_group`, `pinned_subject_fields`) is an explicit per-template toggle; this one isn't, because it's not a UCM-specific policy decision, it's how strong mapping works.
- **Independent of whether the CSR is naked** — the existing AD-derivation code only fires for naked CSRs (GPO autoenrollment), but the SID identifies who authenticated, not what subject was requested, so this applies to manually-typed Kerberos-bound CSRs too (certmgr.msc, certreq).
- **AD Connector not configured/enabled → feature doesn't apply**, issuance proceeds exactly as today. **AD Connector configured but this request's SID lookup fails → issuance is refused.** Worth being explicit about the practical effect: on any install that already uses naked-CSR autoenrollment, this just adds one more lookup to a connection that already has to work. For Kerberos-bound *manual* enrollment specifically, this is a new dependency — today that path never touches AD Connector at all, since only naked CSRs do. `objectSid` is an unprivileged, always-present attribute (not something admins typically restrict), so I don't expect this to bite in practice, but it's a real behavior change worth flagging rather than discovering later.
- UsernamePassword-bound issuance is unaffected — no per-request AD identity exists there, same limitation Enroll ACL already has.

## Ground truth, not spec text

The extension's byte structure was captured from a real certificate issued by a real ADCS Enterprise CA in the lab (not derived from documentation). One non-obvious finding: the SID is embedded as its **string form** (`"S-1-5-21-..."`), not raw binary, despite what the OID name might suggest. Also confirmed empirically that `ldap3` doesn't auto-format AD's `objectSid` — the real binary SID has to be read from `raw_values`, not `.value`.

## Testing

- Unit tests for the SID binary→string conversion and the extension builder, both using the real captured bytes as golden fixtures (byte-for-byte match, not just "doesn't crash").
- End-to-end `issue()` tests covering all the branches above (naked/non-naked Kerberos-bound, AD Connector unconfigured, lookup failure, UsernamePassword-bound).
- Full backend suite: 3508 passed, 6 skipped, 0 failed. (A few existing Kerberos-bound WSTEP tests needed a mock added for the new unconditional lookup — expected, not a sign of anything wrong.)
- Real lab verification: retriggered actual GPO autoenrollment on a domain-joined Windows 11 client, decoded the issued certificate's SID extension, and confirmed it's byte-for-byte the client's real AD SID.

### DIFF
--- a/backend/services/ad_connector/lookup.py
+++ b/backend/services/ad_connector/lookup.py
@@ -221,6 +221,95 @@ def lookup_computer_dns_hostname(sam_account_name):
             pass
 
 
+def _sid_bytes_to_string(raw):
+    """Windows binary SID -> ``"S-1-5-21-...-<RID>"`` string form.
+
+    Layout: ``byte[0]`` revision, ``byte[1]`` sub-authority count,
+    ``bytes[2:8]`` big-endian identifier authority, then N x 4-byte
+    little-endian sub-authorities. Verified against a real captured SID
+    (both a live LDAP query for a lab machine account and the SID string
+    embedded in a real ADCS-issued certificate's security extension agree
+    on the domain-SID prefix) -- not spec-derived.
+
+    ``None`` on any malformed input (wrong length for the declared
+    sub-authority count, empty bytes) rather than raising -- this is
+    untrusted binary data straight from a directory attribute.
+    """
+    if not raw or len(raw) < 8:
+        return None
+    revision = raw[0]
+    sub_count = raw[1]
+    expected_len = 8 + sub_count * 4
+    if len(raw) != expected_len:
+        return None
+    authority = int.from_bytes(raw[2:8], 'big')
+    sub_authorities = [
+        int.from_bytes(raw[8 + i * 4:12 + i * 4], 'little')
+        for i in range(sub_count)
+    ]
+    return f"S-{revision}-{authority}-" + "-".join(str(s) for s in sub_authorities)
+
+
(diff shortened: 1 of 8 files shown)


NeySlim (approved):
True ADCS parity: unconditional SID embedding via the AD Connector, fail-closed when the connector is configured but the SID lookup fails for a specific requester, binary SID parsing is strict, and the encoding is verified byte-for-byte against a real ADCS cert in tests. Suite green (3507 passed). Note: merged after #274 — no interaction (pins never touch the requester-sid path).

PR #281, opened 2026-08-13
feat(scep): Microsoft Intune SCEP challenge validation (#228)

Hemsby:
## Summary

Second half of #228. Intune doesn't support a static SCEP challenge password -- it issues a per-device, per-request encrypted+signed challenge blob that only Intune's own API can validate. A SCEP profile can now be bound to an Entra app registration (tenant ID, client ID, client secret) and validate incoming requests live against Intune instead of comparing a local `challenge_password`.

There's no published REST spec for Intune's SCEP validation API -- Microsoft ships a reference client library instead and documents only its method signatures. I read the actual wire contract out of Microsoft's own MIT-licensed reference implementation ([`microsoft/Intune-Resource-Access`](https://github.com/microsoft/Intune-Resource-Access), `src/CsrValidation/csharp/ScepValidation/`) rather than reconstructing it from the summary docs, so the auth flow, service discovery, and `ScepActions/*` request shapes should match what Microsoft's own client sends call-for-call.

### Design decisions

- **Intune-enabled profiles require `auto_approve=true`.** Intune's flow is a synchronous validate-then-issue round trip -- there's no human approval queue on Intune's side to poll, so I reject the combination at the API layer with a clear error rather than let it silently never work.
- **Notify-before-respond, not best-effort-after.** Re-reading Microsoft's docs, the "if this throws, don't issue" warning on `SendSuccessNotification` only makes sense if "issuance" means handing the cert to the device, not the act of signing it. So the order is: validate -> sign (flushed, not committed) -> notify Intune -> only then commit and return the cert. A notify failure rolls back cleanly -- no orphaned "issued but Intune never heard about it" row, and the device's own SCEP retry re-runs the whole thing against a clean slate.
- **Client secret encryption via `utils.encryption`, not `security.encryption`.** The existing static `challenge_password` field uses `security.encryption.encrypt_text`, which silently stores plaintext if no `/etc/ucm/master.key` is configured. `utils.encryption.encrypt_value` always encrypts (real key or machine-id-derived, never plaintext) -- same choice the AD Connector's bind password already made for the same reason. A real Entra app secret gets the same treatment.

### What's new

- `backend/services/scep/intune_client.py` -- the API client (MSAL client-credentials auth, Graph-based service discovery with cache/clear-on-failure, `validate_request`/`send_success_notification`/`send_failure_notification`)
- `ScepProfile` gains `intune_enabled`/`intune_tenant_id`/`intune_client_id`/`intune_client_secret` (migration 076)
- `POST /api/v2/scep/profiles/test-intune-connection` -- mirrors the AD Connector's inline test-before-save endpoint (token acquisition + service discovery only, never touches `ScepActions/*` so it doesn't spend a real Intune challenge)
- Frontend: Intune section on the SCEP profile form (tenant/client/secret fields, Test Connection button, forced auto-approve, mutual exclusion with the static challenge field), i18n'd across all 9 locales
- Help content updated across all 9 locales: Entra app registration steps on the SCEP guide, and split the old combined "Microsoft Intune / JAMF" static-challenge guidance apart, since it's now actively wrong for Intune specifically. Also added the "Profiles" tab to the long-form guide's Tabs list, which had been missing since profiles themselves shipped in v2.201.

## Test plan

- [x] 36 new tests (unit tests for the Intune client's wire format, full `SCEPService` integration tests covering validate-failure/notify-failure-rollback/success paths, profile API tests for the auto-approve constraint and test-connection endpoint)
- [x] Full backend suite run clean alongside them
- [x] The deployed `IntuneScepClient`, running for real, made a live request to `login.microsoftonline.com` and got back a genuine `AADSTS700038` rejection -- confirms the auth request shape is accepted by production Microsoft infrastructure, not just self-consistent with my reading of the reference source
- [x] Full UI clicked through live: profile creation, the auto-approve constraint firing, Test Connection surfacing a real Microsoft error end-to-end
- [ ] **Not yet tested: a real `ScepActions/validateRequest` response from an actual Intune-issued challenge.** Everything above verifies the request shape and the auth handshake, but I don't have Intune tenant access to get past that point and see a real response. If anyone testing this hits a live Intune SCEP profile, that's the one place I'd want eyes on -- the reference source tells me what the request should look like with high confidence, but I've never seen a real response.

## Known limitations

Same as above: the actual `ValidateRequest`/`SendSuccessNotification` response parsing is implemented per the reference source but not yet exercised against a live Intune tenant. Would appreciate it if anyone with real Intune access (cc @stefanelul2000, since you'd offered) could run a real enrollment through this before it's considered fully proven out.


NeySlim:
Thanks for the thorough work — the write-up of the design decisions and the reference-implementation-derived wire format is exactly the right way to tackle an undocumented API.

One process request before review: this PR targets `feature/scep-intune`, which branched off `main` at v2.208, so our CI (backend/frontend tests, lint, security scanning) doesn't run against it and it's out of sync with `dev`. Could you **retarget this PR to `dev`** and rebase accordingly? Once CI is green there we can give it a proper review.

A couple of early notes from a first read:

1. The two unrelated UI fixes (Toast `break-words`, password-reveal toggle on empty preset-secret fields) are fine to keep here, but they'd honestly be cleaner as a small separate PR — they'd merge quickly without waiting for the Intune validation review.
2. The `utils.encryption` vs `security.encryption` choice for the client secret is noted and reasonable, though it introduces a second encryption path for credential material — we'll want to look at that closely in review.
3. Acknowledged that `ScepActions/ValidateRequest`/`SendSuccessNotification` response parsing is unexercised against a live Intune tenant. We won't treat that as a blocker to landing the code, but we'll keep #228 open until someone with real Intune access (as you noted, @stefanelul2000 offered in the issue) confirms an end-to-end enrollment.

Hemsby:
Done:
- Retargeted this PR to `dev` and rebased on current `dev`. CI should now run against it.
- Split the two unrelated UI fixes (Toast `break-words`, password-reveal toggle on empty fields) out into #283, so they can merge independently without waiting on this review.

This PR is now down to just the 2 Intune-specific commits (feature + docs). Ready for review whenever you have time.

NeySlim:
Small process note: editing the base ref doesn't fire our CI (GitHub doesn't emit a check-triggering event for `edited` on `pull_request`, and the rebase push happened on your fork, which our workflows don't observe). I've closed/reopened the PR — checks are running now. Will start the real review once they're green.

PR #282, opened 2026-08-13
fix(search): wire certificate search boxes to backend across paginated pages

### LINKED ISSUE #280: Certificate search only filters current page instead of entire certificate inventory
## Problem

The certificate search functionality appears to filter only the certificates currently displayed on the active page of the certificate list.

When the total number of certificates exceeds the number of entries displayed on a single page, certificates located on other pages cannot be found through the search function.

## Environment

- UCM Version: v2.207

## Steps to Reproduce

1. Open the Certificates page.
2. Ensure there are more certificates than can be displayed on a single page.
3. Navigate to a page that does not contain the target certificate.
4. Enter a search term matching a certificate located on another page.
5. Observe that no results are returned.
6. Navigate to the page containing the certificate.
7. Repeat the search.
8. The certificate is now found.

## Expected Behavior

The search should query the complete certificate inventory and return matching certificates regardless of which page is currently displayed.

## Actual Behavior

The search appears to be performed only against the records loaded in the current page of the table.

## Impact

Finding certificates becomes difficult when the total number of certificates exceeds the page size.

Users may need to manually browse multiple pages before the desired certificate can be located.

## Suggested Improvement

Implement server-side searching/filtering across the complete certificate inventory, or provide a dedicated global search function that searches all certificates.

Thank you for your work on this project!

Hemsby:
## Summary

Fixes #280. Certificates, SSH Certificates, User Certificates, and the Discovery "Discovered" tab all paginate results server-side, but each page's search box was filtering only the currently-loaded page's rows client-side. A match sitting on a different page was invisible until you navigated to that page.

- Wires each page's search input to the backend's `search` query param and resets to page 1 when the term changes, matching the pattern already used by the Audit Logs and Backup pages.
- The Certificates, SSH Certificates, and User Certificates backend endpoints already supported server-side `search` — the fix there is frontend-only.
- The Discovery `list_discovered` endpoint had no `search` param at all, so this adds one (ilike across subject/issuer/target/serial_number, applied before pagination), consistent with the existing pattern in `cert_list.py`.

Other list pages in the app (CAs, CSRs, Templates, SSH CAs, ACME, SCEP, CRL/OCSP, TrustStore, Operations, Policies, Approvals, Users/Groups, RBAC, HSM) were audited for the same pattern and don't have this bug — they fetch their full dataset unpaginated and filter client-side, so search already covers all records there.

## Test plan
- [x] Deployed to test server (ucm2) and manually verified: Certificates page — searched for a certificate that only existed on page 3 while sitting on page 1, confirmed it was found and pagination reset correctly
- [x] Verified SSH Certificates and User Certificates search requests now include `search=` and return 200
- [x] Verified Discovery's `GET /api/v2/discovery?search=...` returns 200 with the new backend param
- [x] No console errors on any of the four pages

### DIFF
--- a/backend/api/v2/discovery.py
+++ b/backend/api/v2/discovery.py
@@ -301,10 +301,12 @@ def list_discovered():
     offset = _safe_int(request.args.get('offset', 0), 0, lo=0)
     profile_id = request.args.get('profile_id', type=int)
     status_list = request.args.getlist('status')
+    search = request.args.get('search', '').strip()
     svc = _get_service()
     items, total = svc.get_all(limit=limit, offset=offset,
                                profile_id=profile_id,
-                               status=status_list if status_list else None)
+                               status=status_list if status_list else None,
+                               search=search or None)
     return success_response(data={'items': items, 'total': total})
 
 
--- a/backend/services/discovery/query.py
+++ b/backend/services/discovery/query.py
@@ -40,7 +40,7 @@ class QueryMixin:
         return {'total': len(certs), 'updated': updated}
 
     def get_all(self, limit: int = 200, offset: int = 0,
-                profile_id: int = None, status=None) -> Tuple[List[Dict], int]:
+                profile_id: int = None, status=None, search: str = None) -> Tuple[List[Dict], int]:
         """Return discovered certificates with pagination. Returns (items, total).
         status can be str or list[str] for multi-select.
         """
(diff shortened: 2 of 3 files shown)


NeySlim:
Thanks for this — bad timing though: a fix for #280 landing on the same three pages (Certificates, SSH Certificates, User Certificates) was just merged to `dev` a few hours ago, so this PR now conflicts and duplicates those parts.

However, your **Discovery** addition is new and worth keeping — `list_discovered` indeed had no `search` param at all. Could you rebase on current `dev` and trim this PR down to the Discovery part only (backend `search` param in `discovery.py`/`query.py` + the page wiring)? That part would merge cleanly.

Nice job auditing the other list pages too — matches the conclusion we reached when fixing #280.

Hemsby:
Rebased on current `dev` and trimmed down to just the Discovery part, as requested. Dropped the Certificates/SSH Certificates/User Certificates changes since #280 already covers those.

Now just:
- `backend/api/v2/discovery.py` / `backend/services/discovery/query.py`: adds `search` param to `list_discovered`/`get_all` (ilike across subject/issuer/target/serial_number, applied before pagination)
- `DiscoveryPage.jsx`: wires the Discovered tab's search box to it via the existing `externalSearch`/`onSearchChange` pattern

Single commit, should merge cleanly.

PR #290, opened 2026-08-16
fix(acme): Allow bare TLD registration in 'Local Domains' & Bugfix for DATA_DIR

gb-123-git:
#### Change Log:

1. Make the label+dot group optional so bare TLDs are accepted. Combined with find_local_domain_ca's existing parent-walking logic, registering a bare TLD (e.g. "local") now covers all subdomains (foo.local, bar.local, *.foo.local) automatically.

2. The Config class in settings.py derived all path attributes (CA_DIR, CERT_DIR, LOG_FILE, etc.) from the module-level DATA_DIR variable but never exposed DATA_DIR itself as a class attribute. Any code referencing Config.DATA_DIR directly would raise AttributeError, and consumers reading app.config['DATA_DIR'] would find no key. Fixed `settings.py` to expose DATA_DIR to other classes.

_____________

_This PR is sponsored by PMGA Tech LLP_

### DIFF
--- a/backend/api/v2/acme_local_domains.py
+++ b/backend/api/v2/acme_local_domains.py
@@ -1,7 +1,6 @@
 """
 ACME Local Domains API Routes
 Manages domain-to-CA mappings for the Local ACME server.
-
 """
 import re
 import logging
--- a/backend/api/v2/acme_local_domains.py
+++ b/backend/api/v2/acme_local_domains.py
@@ -1,6 +1,9 @@
 """
 ACME Local Domains API Routes
 Manages domain-to-CA mappings for the Local ACME server.
+
+Patched: allow bare TLD registration (e.g. "local", "internal") so that
+parent-walking in find_local_domain_ca covers all subdomains of that TLD.
 """
 import re
 import logging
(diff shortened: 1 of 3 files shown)


NeySlim:
Thanks — both changes are functionally sound, and the bare-TLD regex is exactly right (checked the edge cases: `*`, `..com`, `com.` still rejected; `local` / `*.local` accepted, and `find_local_domain_ca`'s parent-walk covers the subdomains as you describe).

A few asks before this can merge:

1. **Trim the diff to the functional change.** Nearly all of the ~130 changed lines are trailing-whitespace edits — they bury the one-line regex change and pollute blame. Please disable format-on-save for these files and re-push with only the real changes.
2. **Drop the "Patched:" comments.** Comments should describe behavior, not narrate the patch — the accepted/rejected examples list in the `_is_valid_domain` docstring is worth keeping, the "Patched:" phrasing isn't. Also please restore the deleted `# Load environment variables FIRST` comment in settings.py — load order matters there and that comment documents it.
3. **Add a test** for `_is_valid_domain` covering the new bare-TLD behavior (accepted: `local`, `*.local`; still rejected: `*`, `..com`, `com.`, single-char TLDs).
4. **Split the `Config.DATA_DIR` change into its own PR.** It's a legitimate fix — without the key, `session_cleanup_task` silently falls back to `/opt/ucm/data`, which is wrong on custom `DATA_DIR` layouts — but it's unrelated to this PR, and the justification needs correcting: nothing raises `AttributeError` today; the real symptom is the silent fallback. A separate PR with that rationale (and no whitespace churn) merges quickly.

The Local Domains improvement itself is welcome — with the diff trimmed down this is an easy merge.

gb-123-git:
@NeySlim : 
>Trim the diff to the functional change. 

This is due to auto format. :(  I'm sure there is a way to 'Ignore whitespaces' in github diff so that you dont see the formatting diff. Do you really need me to correct this ?

>Drop the "Patched:" comments.

done

> Add a test for _is_valid_domain

done

> Split the Config.DATA_DIR change into its own PR. 

Well.. all the commits get added to the PR automatically. For this I will need to remove my changes, wait for you to merge, then send another PR. Or maybe use a different branch, ask you to merge and then sync. Too much work... Just curious Why do we need to split it into 2 PRs by the way? I have made different commits and in release notes you record this, so is really required ?

gb-123-git:
image

@NeySlim:
Please look at 'Hide Whitespace' in the settings

gb-123-git:
@ Neyslim : Completed as per your suggestions.

NeySlim:
Thanks for the quick iteration. To save us both a third round-trip I've landed this on `dev` directly, split as discussed and credited to you in the changelog:

- 019097eab — `feat(acme): accept bare TLDs in local domains (#290)`: your regex, with the anchor switched from `$` to `\Z`. Subtle Python gotcha: `re.match` with `$` still matches before a trailing newline, so your `example.com\n → False` test case would actually have failed. Which brings up the second thing —
- your test file was a standalone script (`run_tests()` + `__main__`), so pytest collected zero tests from it and CI was green without ever running your cases. I converted the full case list (it was a good list — traversal and newline injection included) to `pytest.mark.parametrize`; 37 cases now run in CI.
- 60db4174e — `fix(config): expose DATA_DIR on Config`: kept as a separate commit with the corrected rationale — nothing raises `AttributeError` today; the real symptom was `session_cleanup_task` silently falling back to `/opt/ucm/data` on custom layouts.

Both ship in the next release. Closing this PR since the work is merged — the bare-TLD feature is a nice improvement, thanks. For future PRs: pytest-style test functions (so CI actually runs them) and one concern per PR, and they'll merge fast.

PR #292, opened 2026-08-17
Harden the ACME proxy: ownership binding, RFC 7638 thumbprints, fewer upstream round-trips

gb-123-git:
#### Security

1. **`new-order` must be kid-signed (RFC 8555 §6.2)**. Previously the kid was only required when acme_eab_required was on. verify_jws validates a jwk-signed JWS against its own inline key with no account lookup, so with EAB off a jwk-signed order skipped the account existence check, the account status check and the EAB identifier restrictions — an unregistered or deactivated key could keep ordering certificates. The native new_order in api/acme/acme_api.py has always required the kid unconditionally; the proxy now matches.

2. **Account must be valid to place an order (§7.3.6)**. Was status == 'deactivated'; the inequality check also covers   `revoked`.

3. **Ownership requires a positive match**. `_verify_order_ownership` previously allowed access when it merely found no contradiction, so a half-populated binding (an account row with no stored thumbprint, or a thumbprint that resolved to no account) authorized any verified requester. It now demands a positive match on `account_id` or `client_jwk_thumbprint`, with `_owner_matches_via_account_row` resolving the missing half through the AcmeAccount row on the deny path so pre-existing half-bound orders keep working. Orders with no binding at all (true legacy rows) are served as before.

4. **RFC 7638 thumbprints**. The requester thumbprint was sha256 of the raw JWK dict, which diverges from the stored AcmeAccount.jwk_thumbprint the moment a client includes optional members such as alg, kid or use — turning a legitimate owner into a mismatch. It now uses the same helper that populates the stored value, with a local RFC 7638 implementation (required members only, lexicographic, no whitespace) as fallback.

5. **Exact authz-to-order resolution**. Challenge and authorization lookups matched with a JSON LIKE %url% on upstream_authz_urls and took .first(), so a URL that is a prefix of a tracked one could bind a foreign order. _find_order_by_authz_url now re-verifies the candidate against the decoded URL list.

6. **Authorize before signing anything upstream**. `get_authz, respond_challenge`, `get_order`, `finalize_order` and `get_certificate` resolve the local order and check ownership before the first `_post_with_account`. An unauthorized caller can no longer make the proxy sign a request with the upstream CA account key. `_resolve_challenge_order` does this locally for CAs that nest challenges under their authorization; the upstream `Link rel="up"` header remains the fallback for Let's Encrypt-style disjoint namespaces, and the request fails closed when neither resolves.

7. **DNS automation is confined to the order's own identifiers**. The automation domain comes from the upstream authz response; it is now checked against the order's domains_list before anything is written into the operator's DNS zone.

8. **Bounded, owner-scoped certificate lookup**. The fallback scan in `_find_order_for_certificate` walked every pending proxy order, one signed upstream POST each — any caller could burn the upstream account's rate limit and probe other tenants' orders. It is now restricted to rows the requester could own (plus unbound legacy rows) and capped.

9. **Strict Link parsing**. `_upstream_authz_url_from_link` scans every link-value and accepts only `rel="up"` (quoted or bare, per RFC 8288) instead of first-link-wins, which could latch onto the `rel="index"` entry that many CAs list first.

10. **Payload shape validation**. A non-object JWS payload on new-order/finalize produced a 500; both now return malformed, and identifiers must be an array.

#### Correctness

1. **Background threads no longer mutate shared state**. `_bg_respond_challenge` used `_refresh_account_session()` to rebind the shared service instance's account to the worker's session, racing the request thread and the sibling threads a multi-domain order spawns. It now constructs its own AcmeProxyService from an explicit account_db_id. (`_refresh_account_session` is kept for callers that reuse an instance across app contexts.)

2. **Wildcard handling uses a prefix strip instead of lstrip('*.')**, which strips characters rather than a prefix.

3. **Removed two unreachable duplicate except PermissionError** clauses in authz and challenge.

4. **Extracted `_rewrite_order_urls`, shared by `get_order` and `finalize_order`**.

#### Performance

1. **TTL caches for the upstream directory (5 min)**, the upstream finalize URL and the challenge-to-order map — thread-safe, bounded at 512 entries, every miss falls back to the previous behavior.

2. **Nonce pooling**. Every upstream response carries a fresh Replay-Nonce (§6.5); harvesting them removes one HEAD /new-nonce round-trip per proxied request. Pooled values are single-use and TTL-bounded, and the existing badNonce retry still covers a stale one.

3. **`finalize` skips its extra POST-as-GET** when the finalize URL was captured at `new-order` or a prior `get_order`.

4. **`reset_proxy_caches()` clears all process-level state**, for tests and for in-place directory URL changes.

#### Dead code removed
1. **self.nonces** — assigned in __init__, never read anywhere in the repo.

2. **_urls_share_ca** — orphaned when the _find_order_for_challenge fallback was removed; zero call sites.

#### Behaviour change to be aware of:

A jwk-signed new-order now returns urn:ietf:params:acme:error:malformed ("Account kid required in protected header") instead of being processed. This is what RFC 8555 §6.2 requires and what the native ACME server already enforced; certbot, acme.sh and win-acme all send the kid on new-order.
 

#### Testing

Verified against the existing suite: 
_test_new_order_requires_kid_when_eab_enabled, test_new_order_allows_kid_when_eab_enabled, test_find_order_for_certificate_matches_upstream_cert_url, test_find_order_fast_path_no_upstream_calls, test_fallback_scan_skips_rows_with_url, test_finalize_without_identity_rejected, test_finalize_with_matching_identity_passes_binding, test_rel_up_link_single_prefix, test_dns_cleanup_dispatched_to_background, test_upstream_link_header_not_forwarded, test_link_not_forwarded_on_upstream_error, plus the test_acme_proxy_ownership.py suite._

One test-infrastructure addition is required: an autouse fixture in backend/tests/conftest.py calling reset_proxy_caches(), because three proxy test modules stub the same upstream directory URL with different payloads and the new caches are process-level.

_______

_This PR has been sponsored by PMGA Tech LLP_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,14 +10,33 @@ Starting with v2.48, UCM uses Major.Build versioning (e.g., 2.48, 2.49). Earlier
 
 ## [Unreleased]
 
-### Added
-- API keys: proper revoke/delete lifecycle — revoke an active key, permanently delete a revoked or expired one; expired keys no longer count against the per-user limit; "Last used" now shown in the account page (#291, contributed by @Hemsby)
-- ACME local domains: bare private TLDs (e.g. `local`, `internal`) can now be registered, covering all their subdomains through the existing parent matching (#290, contributed by @gb-123-git)
+> ⚠️ **Upgrade note — in-flight ACME proxy orders placed with a bare JWK must be
+> re-created.** The proxy's `new-order` now requires a kid (RFC 8555 §6.2) and binds
+> every order to an RFC 7638 thumbprint. A **pending** proxy order created before this
+> release through a **jwk-signed** new-order carries only the legacy thumbprint of the
+> raw JWK dict, which no longer equals the value derived from the requesting key — and
+> the account-row reconciliation cannot resolve it either, because that legacy hash
+> never equals a stored account thumbprint. Those orders become inaccessible and must
+> be re-created. ACME automation self-heals (the client simply places a new order on
+> its next run), and already-issued certificates are unaffected — but a client that
+> polls a saved order URL forever will need a nudge.
+
+### Security
+- **ACME proxy `new-order` requires a kid** — a jwk-signed JWS is verified against its own inline key with no account lookup, so accepting one skipped the account existence/status check *and* the per-EAB identifier restrictions whenever EAB was not mandatory, and left the resulting order with no owner recorded — which the ownership check then served to every other client of that proxy. Identifier and challenge-type validation still runs first, so a client with an unsupported identifier keeps getting `unsupportedIdentifier` rather than an authentication error (#260)
+- **Deactivated *and* revoked accounts are refused on `new-order`** — the status check only covered `deactivated`, so a revoked account could keep ordering certificates
+- **Proxy resources require a positive owner match** — authz, challenge, order and certificate access previously allowed any verified requester when an order's owner binding was half-populated (an account id without a stored thumbprint, or a thumbprint that resolved to no account): the absence of a contradiction was treated as authorization. A match is now required on one of the two fields, and a half-populated binding is reconciled through the owning `AcmeAccount` row rather than waved through (#260)
+- **Owner thumbprints follow RFC 7638** — the thumbprint was computed over the raw JWK dict, so a client sending optional members (`alg`, `kid`, `use`) turned a legitimate owner into a mismatch; only the required members are hashed now, by a single implementation shared with the value stored on the account
+- **Upstream authorizations are matched exactly** — the candidate lookup is a `LIKE '%url%'` prefilter, which also matches a *prefix* of a stored URL (`.../authz-v3/99` inside `.../authz-v3/999`); a client could bind a foreign authorization to an order it does own and pass the ownership check with it. Prefilter hits are now re-checked against the decoded URL list
+- **Challenge-to-order resolution has no approximate fallback** — the owning order is resolved from the recorded authz mapping, from the challenge's parent authz path, or from the upstream `Link rel="up"` header only. There is no host-only match and no "most recent pending order" guess, either of which could run DNS-01 automation against a foreign domain; `rel="up"` is also selected explicitly rather than taking the first link, since `rel="index"` is commonly listed first
+- **The certificate-download fallback scan is owner-scoped** — resolving an untracked certificate URL scanned every pending proxy order and signed one upstream POST per candidate, letting any caller drive the proxy's upstream account (cross-tenant probing, rate-limit burn). The scan is now restricted to rows the requester could own and capped
+- **Ownership is verified before any upstream call** — a denial no longer costs a signed request to the CA
+- **A challenge whose identifier cannot be determined fails closed** — on a multi-SAN order the first domain is not necessarily the one being validated; publishing the DNS-01 TXT record under that name was wrong. The request now fails with a problem document naming the cause
 
 ### Fixed
-- ACME server: http-01 validation now follows HTTP redirects (RFC 8555 §8.3) — a site-wide http→https 301 on the challenge path no longer fails with "Key authorization mismatch". Up to 5 hops, http/https on default ports only, TLS not verified on https hops (the target usually serves the certificate being renewed), and every hop re-vetted by the SSRF policy (cloud metadata always refused; targets resolved, vetted and pinned when private IPs are disallowed)
-- Deleting a certificate (single or bulk) or a CA now removes its cert/key/csr files on disk instead of leaving them orphaned (#289, contributed by @Hemsby)
-- `app.config['DATA_DIR']` is now set, so the session cleanup task no longer silently falls back to the default data directory on custom layouts (#290, contributed by @gb-123-git)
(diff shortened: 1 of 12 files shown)


NeySlim (changes requested):
Thanks — this is a strong hardening pass. I verified the ten security claims against the diff and they hold: positive-match ownership (the old check authorized any verified requester on a half-populated binding), RFC 7638 thumbprints (hashing the raw JWK dict diverged as soon as a client sent optional members), exact authz matching (the `LIKE '%url%'` prefilter could match a prefix of a stored URL), authorization before any upstream signing, DNS automation confined to the order's own identifiers, the bounded owner-scoped certificate scan, strict `Link rel="up"` parsing (the old first-link fallback could latch onto `rel="index"`), and the unconditional kid requirement on new-order (matches the native endpoint, RFC 8555 §6.2). I also confirmed the two removed `except PermissionError` clauses were dead duplicates — each route still has a live handler, so the 403 path is intact.

A few things before this can merge:

**1. CI is red — 7 backend tests fail on the SQLite job.** Two root causes:
- The unconditional kid check now runs before identifier validation, so `test_proxy_rejects_ip_and_tls_alpn_requests` gets `Account kid required in protected header` instead of `unsupportedIdentifier`/`malformed`. Either kid-sign those test payloads or validate the payload shape before the account checks — both are defensible, but the tests must encode the chosen contract.
- `_bg_respond_challenge` now builds its own `AcmeProxyService`, which escapes the mocks the tests patch onto the shared instance — the suite ends up making real network calls and burning 60 s connect timeouts (`test_bg_thread_*`). Please add an injection point (e.g. a small `_make_worker_service()` factory the tests can patch) and adapt the mocks.
- The remaining failures (`test_untracked_certificate_not_found`, `test_proxy_new_order_forwards_replaces_when_present`, `test_sign_and_post_uses_pinned_safe_request_post`) look like call-count/behavior drift from the nonce pooling and the owner-scoped scan — same treatment: update them to the new contract deliberately.

**2. No new tests.** For ten security fixes I'd want regression coverage at least for: positive ownership match across a half-populated binding, thumbprint equality when the request JWK carries `alg`/`use`, the LIKE-prefix authz case, and a `Link` header where `rel="index"` is listed first.

**3. CHANGELOG.** Please add an entry under `[Unreleased]`, including an upgrade note: pending proxy orders created before this change through a jwk-signed new-order are bound only by the legacy raw-dict thumbprint; after this change that value no longer matches the requester (and the account-row reconciliation cannot resolve it, since the legacy hash never equals a stored account thumbprint), so those in-flight orders become inaccessible and must be re-created. Self-healing for ACME automation, but worth stating.

Minor, non-blocking:
- `_jwk_thumbprint` instantiates `AcmeService()` per call and keeps a duplicate RFC 7638 implementation as fallback — I'd use the local `_rfc7638_thumbprint` unconditionally to avoid drift between the two.
- The "a header JWK (if any) wins" comment in `new_order` is moot: jwk and kid are mutually exclusive in a JWS and new-order now requires kid, so `jwk` is always None there.
- `_challenge_domain` raising `RuntimeError` for a multi-SAN order with no authz URL surfaces as a 500; fail-closed is right, but mapping it to a proper problem document would be cleaner.

The cache design (TTL-bounded, size-capped, lock-held, advisory with upstream fallback) and the single-use nonce pool look right, including the autouse reset fixture.

gb-123-git:
@ NeySlim - Please review this thoroughly since I was forced to change some of the test suite files.

NeySlim (approved):
Re-reviewed the update — everything I asked for is in:

- CI is green on both backends; the `_make_worker_service()` injection point cleanly fixes the mock escape, and keeping identifier validation ahead of the kid check is a sound contract (documented in the CHANGELOG).
- The new regression suite covers the half-populated binding, thumbprint stability with optional JWK members, the LIKE-prefix authz case, and `rel="up"` selection with `rel="index"` listed first — plus the multi-SAN fail-closed path.
- CHANGELOG entry with the upgrade note for in-flight jwk-signed orders.
- The minor point is addressed too: `_rfc7638_thumbprint` is now the single implementation.

The SSH deletion guard also checks out: not-found still returns 404 (handled before the service call), a valid non-revoked cert now returns 409, and all three paths are tested.

Merging — thanks for the thorough follow-up!

Hemsby:
Heads up — this merge clobbered the `[2.212]` changelog section. The 3-way merge resolved cleanly (no conflict markers) because #292 rewrote the whole `### Added`/`### Fixed` body while `dev` had only renamed `## [Unreleased]` → `## [2.212] - 2026-08-17` in between, so git combined "new header" + "new body" without flagging it. Result: the entries for #291 (API keys lifecycle), #290 (bare TLD ACME domains + DATA_DIR fix), #289 (orphaned cert file cleanup), and the http-01 redirect-following fix are gone from `dev`'s CHANGELOG.md right now.

NeySlim:
Good catch, thanks — exactly what happened: the PR rewrote the section body while dev had only renamed the header, so the 3-way merge combined them without a conflict. Restored in 4f9195a6b: the #292 content is back under `[Unreleased]` and the full `[2.212]` section (#289, #290, #291, http-01 redirects) is reinstated.

PR #295, opened 2026-08-19
fix azure_provider.py HSM option

jeancarlor:
https://learn.microsoft.com/en-us/python/api/azure-keyvault-keys/azure.keyvault.keys.aio.keyclient?view=azure-python

### DIFF
--- a/backend/services/hsm/azure_provider.py
+++ b/backend/services/hsm/azure_provider.py
@@ -269,7 +269,7 @@ class AzureKeyVaultProvider(BaseHsmProvider):
                 key = self._key_client.create_rsa_key(
                     name=label,
                     size=param,
-                    hsm=True,  # Always use HSM-backed
+                    hardware_protected=True,  # Always use HSM-backed
                     key_operations=key_ops,
                     exportable=extractable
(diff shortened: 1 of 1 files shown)


NeySlim:
Thanks for the fix — confirmed: `azure-keyvault-keys` takes `hardware_protected`, not `hsm`, so `create_rsa_key`/`create_ec_key` raise a `TypeError` as-is and Azure-backed key generation is broken without this.

One thing before it can go in: all changes land on `dev` first and ride the normal release train to `main`. Could you change this PR's base branch to `dev` (Edit next to the title → base: `dev`)? The file is identical on both branches, so it should retarget cleanly with no rebase needed.

jeancarlor:
I changed the PR to the dev branch.

NeySlim (approved):
Verified: `azure-keyvault-keys` (>=4.9.0 pinned in requirements) has no `hsm` parameter on `create_rsa_key`/`create_ec_key` — `hardware_protected` is the documented keyword, so this unbreaks Azure HSM-backed key generation. CI green on both backends. Thanks!

PR #297, opened 2026-08-21
fix (security) : complete re-work of certificate renewal cycle

gb-123-git:
# UCM Enhancement: Rework Certificate Renewal Lifecyle; Revoke & Delete - Persistent Revocation; Delete Guard

## Overview

This patch implements interrelated features for the Ultimate Certificate Manager (UCM):

1. **Deletion guard** — prevents deletion of valid (non-revoked, non-expired) certificates without prior revocation.
2. **Persistent revocation records** — a new `revoked_serials` table stores revocation metadata independently of the `certificates` table, so revocation data survives certificate deletion and continues to appear in CRLs and OCSP responses until the original certificate expires.
3. **Renewal rework** — a shared `renew_certificate_in_place` function (`services/cert/renewal.py`) is the single implementation used by all three renewal callers (manual, bulk, auto). It records the old serial in `revoked_serials` (reason: `superseded`) with `certificate_id` preserved, then updates the existing certificate row in-place with a new serial number, key, and validity. The certificate's `id`, `refid`, and `created_at` are preserved. `renewed_at` (DateTime) and `renewed_times` (Integer) columns track the last renewal timestamp and count.
4. **Auto-renewal rewrite** — `AutoRenewalService` now delegates to the same shared renewal function with `rekey=False` (protocol-enrolled certs keep their client-side key). The old create-new-row + archive-old-row pattern is eliminated, giving unified renewal semantics across manual, bulk, and automatic renewal.
5. **Private key encryption at rest** — renewed private keys are encrypted with `encrypt_private_key` exactly as at issuance, preventing a silent downgrade to plaintext on deployments with `master.key` configured.
6. **Revoked certificate renewal guard** — `renew_certificate_in_place` rejects revoked certificates with HTTP 409, preventing a renewal from silently clearing revocation state.
7. **Atomic revocation persistence** — `revoke_certificate` now flushes the certificate revocation and the `RevokedSerial` record in a single transaction, so a worker can never observe a revoked cert without its persistent record.
8. **Type-safe JSON config parsing** — `auto_renewal_service.py` validates that `auto_renewal_sources` and `auto_renewal_notify_emails` parse to lists, falling back to defaults instead of silently iterating characters.
9. **Atomic unhold** — `unhold_certificate` now clears the cert revocation and deletes the `RevokedSerial` record in a single transaction, so a worker can never observe an unheld cert whose serial is still on the CRL.

---

### Deletion without revocation
Certificates could be deleted without being revoked first. A deleted valid certificate disappears from the database silently — no audit trail, no CRL entry, no OCSP status change. Clients relying on the cert have no way to know it was administratively removed.

### Revocation lost on deletion
When a revoked certificate was deleted (e.g. after renewal), its row in the `certificates` table was removed. Since CRL generation and OCSP both queried the `certificates` table for `revoked=True` rows, the revocation would vanish from subsequent CRLs and OCSP responses. This is a PKI correctness issue — a revoked serial should remain on the CRL until the certificate's `notAfter` has passed.

### Renewal overwrote in-place without revoking the old serial
The renewal endpoint replaced the certificate's `crt`, `prv`, `serial_number`, `valid_from`, and `valid_to` fields on the same database row, and cleared `revoked=False`. The old serial number was never revoked, so it never appeared on the CRL. A client checking the old serial via OCSP would see "good" (if the row was already overwritten) or "unknown" (if deleted).

### Auto-renewal used a different renewal model
`AutoRenewalService.renew_certificate` created a *new* `Certificate` row and set `archived = True` on the old one. It never revoked the old serial and never wrote a `RevokedSerial`. SCEP/ACME/EST auto-renewals left the superseded serial reporting **good** via OCSP forever, while manual/bulk renewal marked it `superseded`. Two incompatible renewal semantics coexisted.

### Renewal wrote private keys unencrypted
Issuance encrypted private keys with `encrypt_private_key`, but renewal stored the new key in plaintext. On a deployment with `master.key` configured, every renewal silently downgraded that cert's key to plaintext at rest. `decrypt_private_key` passes plaintext through, so nothing broke functionally — the failure was silent.

---

### 1. Deletion Guard

**Files:**
- `backend/api/v2/certificates/cert_lifecycle.py`
- `backend/api/v2/certificates/bulk.py`

**Changes:**
- `delete_certificate` endpoint: before calling `CertificateService.delete_certificate`, checks if the cert is valid (`not revoked` and `valid_to > now`). If so, returns **HTTP 409 Conflict** with message `"Cannot delete a valid certificate. Revoke it first."`.
- `bulk_delete_certificates` endpoint: same check per certificate. Valid certs are added to the `failed` list with the error message instead of being deleted.
- Revoked and expired certificates can still be deleted normally.

### 2. RevokedSerial Model & Migration

**Files:**
- `backend/models/revoked_serial.py` (new)
- `backend/migrations/078_revoked_serials.py` (new)
- `backend/models/__init__.py` (modified — import + `__all__`)

**RevokedSerial model fields:**

| Field | Type | Description |
|-------|------|-------------|
| `id` | Integer PK | Auto-increment primary key |
| `caref` | String(36) | CA reference ID (FK to `certificate_authorities.refid`) |
| `serial_number` | String(100) | Certificate serial number (hex string, indexed) |
| `revoked_at` | DateTime | When the certificate was revoked |
| `revoke_reason` | String(100) | RFC 5280 revocation reason (e.g. `key_compromise`, `superseded`) |
| `invalidity_at` | DateTime | RFC 5280 §5.3.2 invalidityDate (optional) |
| `valid_to` | DateTime | The original certificate's notAfter — used to determine when the entry is stale |
| `created_at` | DateTime | Record creation timestamp |
| `certificate_id` | Integer | Optional FK to the certificate row (nulled when the cert is deleted) |

**Migration 078:**
- Multi-backend (SQLite + PostgreSQL compatible)
- Creates `revoked_serials` table with indexes on `caref`, `serial_number`, and `certificate_id`
- Indexes are created unconditionally (`CREATE INDEX IF NOT EXISTS`) outside the table-existence check, so they are added even when the table pre-exists without them
- Uses `conn.execute()` for the single CREATE TABLE DDL instead of `conn.executescript()` (single statement does not require script execution)
- Idempotent — skips table creation if the table already exists, but always ensures indexes exist

### 3. Revocation Inserts Persistent Record

**File:** `backend/services/cert/mixins/lifecycle.py`

**Changes in `revoke_certificate`:**
- Sets `revoked=True` on the certificate, then **flushes** (not commits) so the in-memory state is visible to the `RevokedSerial` query within the same transaction.
- Checks if a `RevokedSerial` entry already exists for the same `caref` + `serial_number`.
- **Duplicate entry exists** — updates the existing row in-place with the latest `revoked_at`, `revoke_reason`, `invalidity_at`, `valid_to`, and `certificate_id` (so a hold → unhold → re-revoke cycle carries the latest reason on the CRL).
- **No duplicate** — inserts a new `RevokedSerial` record with the cert's `caref`, `serial_number`, `revoked_at`, `revoke_reason`, `invalidity_at`, `valid_to`, and `certificate_id`.
- **Single atomic commit** — the certificate revocation and the `RevokedSerial` record are persisted together in one `db.session.commit()`. A worker can never observe a revoked cert without its persistent record.
- **Commit fails** — the entire revocation is rolled back via `_restore_revocation_state`, which restores the exact pre-revocation values (`revoked`, `revoked_at`, `revoke_reason`, `invalidity_at`) and commits. A `RuntimeError` is raised so the caller knows the revocation did not take effect. The certificate must not appear revoked if the revocation record cannot be persisted.

**Changes in `delete_certificate`:**
- Before deleting the certificate row, nulls out `RevokedSerial.certificate_id` for any rows pointing to this cert. This ensures the CRL/OCSP query treats them as orphaned (the cert row is gone, but the revocation data persists).

### 4. CRL Generation Uses RevokedSerial + Purge

**Files:**
- `backend/services/crl/query.py`
- `backend/services/crl/generation.py`

**Changes in `CRLQueryMixin.get_revoked_certificates`:**
- Returns a merged list of:
  1. **Live certs** — `Certificate` rows with `revoked=True` AND `valid_to > now` (expired certs excluded per RFC 5280)
  2. **Orphaned revocations** — `RevokedSerial` rows with `valid_to > now` whose `serial_number` is not in the live set AND whose `certificate_id` is NULL or points to a non-revoked cert
- Deduplicates by `serial_number` — live `Certificate` rows always take precedence
- Expired certs are excluded from both sources — clients reject them on validity alone, so they don't need CRL entries

**New method `CRLQueryMixin.purge_stale_revoked_serials`:**
- Deletes `RevokedSerial` entries where `valid_to < utc_now()`
- Called **before** `get_revoked_certificates` during each full CRL generation, but only when `is_purge_stale_serials_enabled()` returns `True`
- Gated by the `crl_auto_purge_stale_serials` system config key — defaults to `False` (off) so entries are preserved as audit records (renewal chain history) unless an admin explicitly enables purging
- Returns the count of purged entries
- Keeps the `revoked_serials` table clean — once a certificate has expired, its serial no longer needs to appear on the CRL

**New method `CRLQueryMixin.is_auto_delete_enabled`:**
- Reads the `crl_auto_delete_expired_revoked` system config key from the `system_config` table
- Defaults to `false` (off) — expired revoked certs remain in the database as historical records
- Admin can enable via the UCM settings UI (Security → HSTS & OCSP section) or by inserting/updating the config row

**New method `CRLQueryMixin.purge_expired_revoked_certificates`:**
- Deletes `Certificate` rows that are **both** revoked AND past `valid_to`
- Nulls out `RevokedSerial.certificate_id` FK before deleting (same pattern as `delete_certificate`)
- Safe — the `RevokedSerial` table persists the revocation data independently
- Only called when `is_auto_delete_enabled()` returns `True`
- Returns the number of deleted certificate rows

**Changes in `generate_crl` (full CRL):**
- If `is_purge_stale_serials_enabled()` returns `True`, calls `purge_stale_revoked_serials(ca_id)` before building the CRL
- If `is_auto_delete_enabled()` returns `True`, calls `purge_expired_revoked_certificates(ca_id)` after purge and before `get_revoked_certificates`

**Changes in `generate_delta_crl`:**
- Also queries `RevokedSerial` for entries revoked after the base CRL's `this_update`
- Merges orphaned entries into the delta CRL (deduplicated against live certs)

### 5. OCSP Fallback to RevokedSerial

**File:** `backend/services/ocsp_service.py`

**Changes in `_status_for_serial`:**
- When a `Certificate` row is not found for a given serial, falls back to querying `RevokedSerial`.
- If a `RevokedSerial` entry exists, returns status `revoked` with the recorded `revoked_at` and `revoke_reason`.
- If no `RevokedSerial` entry exists either, returns `unknown` as before.

**Changes in `generate_response`:**
- Same fallback logic — when the certificate row is not found, checks `RevokedSerial` before returning `unknown`.

This ensures that OCSP clients querying a deleted-but-revoked certificate's serial receive `revoked` status instead of `unknown`.

### 6. Shared Renewal Function (`renewal.py`)

**File:** `backend/services/cert/renewal.py` (new)

This module is the single implementation of in-place certificate renewal, called by all three renewal paths:

- `POST /api/v2/certificates/<id>/renew` — manual renewal (`cert_renew.py`)
- `POST /api/v2/certificates/bulk/renew` — bulk renewal (`bulk.py`)
- `AutoRenewalService.renew_certificate` — automatic renewal (`auto_renewal_service.py`)

**`renew_certificate_in_place(cert, ca=None, *, username, actor_user_id, rekey, regenerate_crl, trigger)`**

1. **Resolve the issuing CA** by refid → subject → CN fallback.
2. **Validate** — cert has data, not revoked (409), not MSCA-sourced, CA exists, CA has private key, CA is not offline, CA cert not expired.
3. **Key strategy:**
   - `rekey=True` — generate a fresh key pair matching the original's algorithm and size (manual + bulk renewal).
   - `rekey=False` — re-sign the existing public key (auto-renewal for SCEP/EST/ACME where the private key lives on the client).
4. **Re-validate NameConstraints** against the CA chain before re-issuing (RFC 5280 §4.2.1.10), with `renewal_of` granting renewal-at-par for names the certificate already carries.
5. **Build and sign** the new certificate — same duration as the original, clamped to 1..3650 days and to the CA's own expiry. Original extensions are carried across; SKI/AKI are regenerated.
6. **Stage the superseded serial** via `_record_superseded_serial` — inserts or updates a `RevokedSerial` entry with reason `superseded`, `certificate_id` preserved (not nulled), and `valid_to` from the old cert.
7. **Update the `Certificate` row in-place:**
   - New `crt` (base64-encoded PEM), `serial_number`, `aki`, `ski`, `valid_from`, `valid_to`, `key_algo`, `issuer`, SANs
   - `prv` — encrypted with `encrypt_private_key` when `rekey=True` (matches issuance in `lifecycle.py`, preventing a silent downgrade to plaintext at rest)
   - `revoked = False`, `revoked_at = None`, `revoke_reason = None`, `invalidity_at = None`, `archived = False`
   - `renewed_at = utc_now()`, `renewed_times` incremented by 1
   - **Preserved:** `id`, `refid`, `created_at`, `created_by`, `descr`, `caref`, `cert_type`, `subject`, `ocsp_uri`, `ocsp_must_staple`, `private_key_location`, `source`, `template_id`, `template_overrides`
8. **Atomic commit** — RevokedSerial + in-place cert update commit together via `commit_or_rollback`. On failure both roll back; no partial state.
9. **Overwrite cert/key files on disk** (filenames are refid-based, unchanged).
10. **Invalidate OCSP cache** for the old serial.
11. **Regenerate CRL** if `regenerate_crl=True` and CA has CDP enabled (bulk and auto callers pass `False` and regenerate once per CA after the batch).
12. **Audit log** — `certificate_renewed` action with trigger, rekey status, old and new serials.
13. **Webhook** — `emit_cert_renewed` fires with the updated cert data.

**Why `certificate_id` is preserved on the `RevokedSerial` entry:**
The CRL query (`_filter_orphan_serials`) checks whether the `certificate_id` still points to a live, non-revoked cert carrying the same serial. If the serials differ (cert was renewed in-place), the old serial stays on the CRL. This preserves a direct FK link from every previous serial back to the certificate row, enabling `RevokedSerial.filter_by(certificate_id=X)` to retrieve the complete renewal chain.

**MSCA renewal path:** `cert_renew.py` delegates to `_renew_msca_certificate` for MSCA-sourced certs — out of scope for the shared function.

**`RenewalError`** — raised with an HTTP `status` field for API callers; service callers use `message` only. The session is left clean (rolled back) in every failure path.

### 7. Manual Renewal Route

**File:** `backend/api/v2/certificates/cert_renew.py`

Calls `renew_certificate_in_place(cert, rekey=True, regenerate_crl=True, trigger='manual')`. MSCA-sourced certs are routed to `_renew_msca_certificate` before reaching the shared function.

### 8. Bulk Renewal Route

**File:** `backend/api/v2/certificates/bulk.py`

Iterates through certificate IDs and calls `renew_certificate_in_place(cert, rekey=True, regenerate_crl=False, trigger='bulk')` per cert. CRL regeneration is deferred — `renewed_ca_ids` is collected during the loop, then `CRLService.generate_crl` is called once per CDP-enabled CA after the loop completes. The `results['success']` list returns bare cert IDs (IDs don't change with in-place renewal).

### 9. Auto-Renewal Service

**File:** `backend/services/auto_renewal_service.py`

**Previous behavior:** Created a new `Certificate` row, set `archived=True` on the old one, never revoked the old serial, never wrote a `RevokedSerial`. Two incompatible renewal semantics coexisted (manual/bulk vs auto).

**New behavior:** `renew_certificate(cert, regenerate_crl=False)` calls `renew_certificate_in_place(cert, rekey=False, trigger='auto')`. Auto-renewed certificates get identical treatment to manual/bulk: superseded serial on the CRL, in-place row update, `renewed_at`/`renewed_times` maintained, audit entry, webhook.

`run_auto_renewal()` iterates eligible certs, calls `renew_certificate` per cert with `regenerate_crl=False`, then calls `_publish_crls(renewed_carefs)` once per CDP-enabled CA after the batch.

**Key strategy (`rekey=False`):** Protocol-enrolled certificates (SCEP/EST/ACME) keep their client-side private key — UCM only ever saw the public half, so issuing a new key pair would hand the device a certificate it cannot use. The existing public key is re-signed instead.

**Eligibility:** Certificates with `revoked=False`, `archived=False`, `valid_to` within the configured threshold, `crt` present, and `source` in the configured renewal sources list.

**Failure handling:** `_log_failure` writes an audit entry (`certificate.auto_renewal_failed`) so failures are visible. Notifications are sent via email if configured.

### 10. CRL Auto-Delete & Auto-Purge UI

**Files:**
- `backend/api/v2/settings/general.py` (modified)
- `frontend/src/pages/settings/SecuritySection.jsx` (modified)
- `frontend/src/i18n/locales/*.json` (9 files modified — en, de, es, fr, it, ja, pt, uk, zh)

**Backend (`general.py`):**
- Added `'crl_auto_delete_expired_revoked'` and `'crl_auto_purge_stale_serials'` to `_ADMIN_ONLY_SETTINGS` frozenset — requires `admin:settings` permission to modify
- GET `/api/v2/settings/general` now returns both settings as booleans (reads from `SystemConfig`, defaults to `'false'`)
- PATCH `/api/v2/settings/general` now accepts both settings in `allowed_keys` — boolean values are converted to `'true'`/`'false'` strings before `set_config`

**Frontend (`SecuritySection.jsx`):**
- Added a `ToggleSwitch` for `crl_auto_delete_expired_revoked` in the HSTS & OCSP settings section, placed between the OCSP response validity input and the save button
- Added a `ToggleSwitch` for `crl_auto_purge_stale_serials` directly below the auto-delete toggle
- Both toggles are disabled for non-admin users (requires `admin:settings` permission)
- Uses i18n keys `settings.crlAutoDeleteExpiredRevoked` (label) + `settings.crlAutoDeleteExpiredRevokedDesc` (description) and `settings.crlAutoPurgeStaleSerials` (label) + `settings.crlAutoPurgeStaleSerialsDesc` (description)

**i18n:**
- Added `crlAutoDeleteExpiredRevoked`, `crlAutoDeleteExpiredRevokedDesc`, `crlAutoPurgeStaleSerials`, and `crlAutoPurgeStaleSerialsDesc` translation keys to all 9 locale files

### 11. Tests

**File:** `backend/tests/test_certificates.py`

**Replaced tests:**
- `test_delete_existing` → `test_delete_valid_cert_blocked` — 409 on deleting a valid (non-revoked, non-expired) cert
- `test_renew_clears_revocation` → `test_renew_revoked_cert_blocked` — asserts 409 when renewing a revoked cert (revoked certs must not be renewed; issue a new cert instead)
- `test_bulk_delete_valid` → `test_bulk_delete_valid_certs_blocked` — valid certs blocked in bulk delete

**Modified tests:**
- `test_delete_confirms_gone` — revokes cert first (deletion guard blocks valid certs)
- `test_delete_service_failure_returns_500` — revokes cert first (deletion guard blocks valid certs)
- `test_bulk_delete_confirms_gone` — revokes cert first (deletion guard blocks valid certs)

**New tests:**
- `TestDeleteCertificate.test_delete_revoked_cert_allowed` — revoked certs can be deleted
- `TestDeleteCertificate.test_delete_expired_cert_allowed` — expired certs can be deleted
- `TestRenewCertificate.test_renew_preserves_row_id` — verifies id/refid preserved after in-place renewal
- `TestRenewCertificate.test_renew_sets_renewed_at` — verifies `renewed_at` set and `renewed_times` increments (0→1→2)
- `TestRenewCertificate.test_renew_preserves_created_at` — verifies `created_at` unchanged across renewal
- `TestRenewCertificate.test_renew_revokes_old_cert` — verifies `RevokedSerial` entry exists for old serial with reason `superseded`
- `TestRenewCertificate.test_renew_revoked_cert_blocked` — asserts 409 when renewing a revoked cert
- `TestBulkDelete.test_bulk_delete_revoked_allowed` — revoked certs allowed in bulk delete
- `TestBulkDelete.test_bulk_delete_mixed` — mixed valid/revoked scenario
- `TestRevokedSerialPersistence.test_revoke_creates_revoked_serial` — verifies revoking a cert inserts a `RevokedSerial` record with correct reason and `valid_to`
- `TestRevokedSerialPersistence.test_revoked_serial_survives_deletion` — verifies that after deleting a revoked cert, the `RevokedSerial` record persists
- `TestRevokedSerialPersistence.test_stale_revoked_serial_purged` — verifies that `RevokedSerial` entries past `valid_to` are purged during CRL generation (enables `crl_auto_purge_stale_serials` via `SystemConfig` first, since the purge is gated by this setting)

---

## Files Changed

| File | Status | Description |
|------|--------|-------------|
| `backend/api/v2/cas/crud.py` | Modified | `delete_ca`: clean up `RevokedSerial` rows for the CA's refid (FK violation fix for PostgreSQL) |
| `backend/api/v2/certificates/bulk.py` | Modified | Bulk delete guard for valid certs; bulk renew calls `renew_certificate_in_place(rekey=True, regenerate_crl=False)` per cert; CRL regenerated once per CA after the loop |
| `backend/api/v2/certificates/cert_lifecycle.py` | Modified | Single delete guard for valid certs (409 on non-revoked, non-expired); unhold removes `RevokedSerial` entry atomically with cert unhold in a single transaction |
| `backend/api/v2/certificates/cert_renew.py` | Modified | Manual renewal calls `renew_certificate_in_place(rekey=True, regenerate_crl=True, trigger='manual')`; MSCA certs routed to `_renew_msca_certificate` |
| `backend/api/v2/settings/general.py` | Modified | Expose `crl_auto_delete_expired_revoked` and `crl_auto_purge_stale_serials` in GET/PATCH settings API; both added to `_ADMIN_ONLY_SETTINGS` |
| `backend/migrations/078_revoked_serials.py` | Modified | Migration for `revoked_serials` table + `renewed_at`/`renewed_times` columns; `certificate_id` index added; indexes created unconditionally; `conn.execute` instead of `conn.executescript` |
| `backend/models/certificate.py` | Modified | `Certificate` model with `renewed_at` and `renewed_times` columns |
| `backend/models/revoked_serial.py` | **New** | `RevokedSerial` model — persistent revocation record that survives certificate deletion |
| `backend/models/__init__.py` | Modified | Import + register `RevokedSerial` in `__all__` |
| `backend/services/auto_renewal_service.py` | Modified | Rewritten to call `renew_certificate_in_place(rekey=False, trigger='auto')`; CRL regenerated once per CA via `_publish_crls` after batch; old create-new-row + archive pattern eliminated; type-safe JSON config parsing (`_safe_json_loads` + `isinstance` guards for `renewal_sources` and `notify_emails`) |
| `backend/services/cert/mixins/lifecycle.py` | Modified | Insert/update `RevokedSerial` on revoke (duplicate-aware, atomic flush + single commit, `_restore_revocation_state` rollback on failure); null FK on delete; detach ACME order FKs on delete; `_suppress_events` parameter on revoke/delete |
| `backend/services/cert/renewal.py` | **New** | Shared `renew_certificate_in_place` function — single implementation for manual, bulk, and auto renewal; `encrypt_private_key` on renewed keys; `RenewalError` with HTTP status; revoked-cert guard (409); NameConstraints re-validation; atomic commit via `commit_or_rollback` |
| `backend/services/crl/generation.py` | Modified | Purge stale entries before query (gated by `crl_auto_purge_stale_serials`); auto-delete expired revoked certs (gated by `crl_auto_delete_expired_revoked`); delta CRL includes orphaned `RevokedSerial` entries |
| `backend/services/crl/query.py` | Modified | Merged query (live + orphaned, `valid_to` filtered); `_filter_orphan_serials` shared by full and delta CRL; `purge_stale_revoked_serials`; `is_purge_stale_serials_enabled()` + `is_auto_delete_enabled()` setting checks |
| `backend/services/ocsp_service.py` | Modified | Fallback to `RevokedSerial` for deleted certs in `_status_for_serial` and `generate_response` |
| `backend/tests/test_certificates.py` | Modified | New + updated tests for renewal and revocation; `test_stale_revoked_serial_purged` enables `crl_auto_purge_stale_serials` config before CRL generation |
| `frontend/src/pages/settings/SecuritySection.jsx` | Modified | ToggleSwitch for `crl_auto_delete_expired_revoked` and `crl_auto_purge_stale_serials` in HSTS & OCSP section |
| `frontend/src/i18n/locales/*.json` | Modified | Added `crlAutoDeleteExpiredRevoked` + `crlAutoDeleteExpiredRevokedDesc` + `crlAutoPurgeStaleSerials` + `crlAutoPurgeStaleSerialsDesc` to all 9 locale files (en, de, es, fr, it, ja, pt, uk, zh) |

---

## Data Flow

```
Revocation:
  revoke_certificate()
    → sets cert.revoked = True, flushes (not commits)
    → inserts or updates RevokedSerial(caref, serial, revoked_at, reason, valid_to, certificate_id)
    → single atomic commit (cert revocation + RevokedSerial together)
    → if commit fails: _restore_revocation_state rolls back cert to pre-revocation state
    → auto-generates CRL if CDP enabled
    → invalidates OCSP cache

Deletion (of revoked cert):
  delete_certificate()
    → nulls RevokedSerial.certificate_id
    → detaches AcmeClientOrder.certificate_id, .source_certificate_id, AcmeOrder.certificate_id
    → deletes cert row + files
    → RevokedSerial record persists

Unhold (lift certificateHold):
  unhold_certificate()
    → if delta CRL enabled: stage removeFromCRL reason, commit, generate delta CRL
    → clears cert.revoked/revoked_at/revoke_reason/invalidity_at
    → deletes RevokedSerial for this caref + serial_number
    → single atomic commit (cert unhold + RevokedSerial deletion together)
    → if commit fails: restores pre-unhold state, returns 500
    → regenerates full CRL, invalidates OCSP cache

CRL Generation:
  if purge_stale_enabled:
    purge_stale_revoked_serials()
      → deletes RevokedSerial where valid_to < now
  if auto_delete_enabled:
    purge_expired_revoked_certificates()
      → deletes Certificate rows where revoked=True AND valid_to < now
      → nulls RevokedSerial.certificate_id FK first
  get_revoked_certificates()
    → live: Certificate.revoked=True AND valid_to > now
    → orphaned: RevokedSerial valid_to > now (no live cert row)
    → dedup by serial_number

OCSP:
  _status_for_serial()
    → Certificate found? → check revoked flag
    → Certificate not found? → check RevokedSerial
    → RevokedSerial found? → return "revoked"
    → neither found? → return "unknown"

Renewal (shared function — all three callers):
  renew_certificate_in_place(cert, rekey=?, regenerate_crl=?, trigger=?)
    → resolve issuing CA (refid → subject → CN fallback)
    → validate: not MSCA, CA exists, CA has key, CA not offline, CA not expired
    → generate new key (rekey=True) or reuse public key (rekey=False)
    → re-validate NameConstraints (renewal_at_par)
    → build + sign new certificate (same duration, clamped to CA expiry)
    → _record_superseded_serial(old serial → RevokedSerial, reason: superseded)
    → update Certificate row in-place (new serial, crt, prv encrypted, validity, renewed_at, renewed_times)
    → commit_or_rollback (RevokedSerial + in-place update atomic)
    → overwrite cert/key files on disk (refid-based filenames unchanged)
    → invalidate OCSP cache for old serial
    → if regenerate_crl and CDP enabled: generate CRL
    → audit log (certificate_renewed, trigger, rekey status, old+new serials)
    → webhook (cert_renewed)

  Callers:
    Manual:  cert_renew.py     → rekey=True,  regenerate_crl=True,  trigger='manual'
    Bulk:    bulk.py            → rekey=True,  regenerate_crl=False, trigger='bulk' (CRL per CA after loop)
    Auto:    auto_renewal.py    → rekey=False, regenerate_crl=False, trigger='auto' (CRL per CA via _publish_crls)
```

---

_This work has been sponsored by PMGA Tech LLP_


NeySlim:
Thanks for this contribution — the core idea is right and it addresses a real PKI correctness gap: a revocation must survive deletion of the certificate row so the serial keeps appearing in CRLs and OCSP responses until the original `notAfter` passes. The `revoked_serials` design, the CRL/OCSP integration, and the dual-backend migration all follow the codebase patterns well. However, the PR cannot be merged in its current state: the renewal path is broken by a Python scoping bug, and there are two PostgreSQL-specific integrity issues plus an atomicity regression that need to be addressed first.

## Blocking

### 1. Every renewal fails with `UnboundLocalError` (this is the red CI)

`cert_renew.py` already imports `ExtensionOID` at module level (line 15) and uses it at line ~122 when copying SANs into the new certificate. The PR adds a **function-local** `from cryptography.x509.oid import ExtensionOID` further down (~line 232), which makes `ExtensionOID` a local variable for the *entire* function scope — so the earlier use at line 122 raises `UnboundLocalError` on every `/renew` call.

This is exactly what the failing SQLite job shows (5 failures: `test_renew_valid_cert`, `test_renew_creates_new_row`, `test_renew_revokes_old_cert`, `test_renew_clears_revocation`, `test_bulk_renew_valid`), reproduced locally. Fix: delete the local import.

Note the green "Backend Tests (PostgreSQL)" job is misleading — it only runs tests marked `postgres` (with `|| true`) plus the migration-runner tests, so it never exercises the renewal path.

### 2. The `caref` FK makes a CA undeletable on PostgreSQL

Migration 078 creates `caref REFERENCES certificate_authorities(refid)`, but `delete_ca` (`api/v2/cas/crud.py`) only cleans up `CRLMetadata` and `OCSPResponse`. Common sequence: revoke certs → delete them (now mandatory for valid certs per #296) → delete the CA. The `issued_certs > 0` guard passes (the cert rows are gone), but the `revoked_serials` rows remain → FK violation on PostgreSQL → 500, and the CA stays undeletable until the entries expire and get purged (potentially years). On SQLite, foreign keys aren't enforced (no `PRAGMA foreign_keys`), so the delete "works" and leaves orphaned rows — a silent behavioral divergence between backends.

Fix: delete (or handle) `RevokedSerial` rows for the CA's `refid` in `delete_ca`, alongside the existing CRL/OCSP cleanup. Deleting them is arguably correct — with the CA gone, nobody can serve its CRL/OCSP anymore.

### 3. Renewal has a data-loss window

The new flow is: revoke old (commit) → **delete old (commit, disk files unlinked too)** → build new row → final commit. If that final commit fails, the old certificate and its private key are already destroyed and the new one was never persisted — the API returns 500 and the certificate is simply gone. The previous in-place update was a single commit, hence atomic.

Suggested fix: create and commit the new row **first**, then revoke + delete the old one (both service calls are already tolerant). Also, the return value of `CertificateService.delete_certificate` is not checked in either renewal path — a failed delete currently continues silently.

## Should be addressed

### 4. ACME order FKs are not detached before the delete

`AcmeClientOrder.certificate_id` and `source_certificate_id` are real FKs to `certificates.id`. The ACME prune path detaches them explicitly before deleting a certificate; the new renewal path calls `delete_certificate` without doing so. Renewing an ACME-sourced certificate will hit an `IntegrityError` on PostgreSQL (delete fails, is silently ignored → duplicate rows: old revoked + new), and leaves dangling references on SQLite.

### 5. Event storm and API contract changes

- Each renewal now emits `cert_revoked` + `cert_deleted` + `cert_renewed` through the event bus (webhooks **and** email subscribers), plus three audit entries. Subscribers will receive "revoked" and "deleted" notifications for what is semantically a renewal. This deserves either suppression/grouping in the renewal flow or at least a deliberate, documented decision.
- **Bulk renew: `results['success']` now contains the *new* certificate ids** instead of the submitted ids — any consumer correlating request↔result breaks. Consider returning `{old_id, new_id}` pairs.
- Unlike the create path, the new row never gets its cert/key files written to disk (and the old set was unlinked by the delete) — inconsistent with `create_certificate`.

## Minor

- The "temporary: added debug logging" commit leaks exception details into API responses (`error_response(f'...: {e}')`, `'error': f'Renewal failed: {e}'`). Please drop it before merge — internals belong in logs, responses stay generic.
- If a serial already has a `RevokedSerial` entry (e.g. hold → unhold → later re-revocation), the insert is skipped without updating `revoked_at`/`revoke_reason`, so the CRL will carry the original hold date/reason. Consider updating the existing row instead of skipping; `unhold` should probably remove the entry as well.
- `valid_to or (utc_now() + timedelta(days=365))` — an arbitrary 1-year CRL residency for certs without `valid_to`; worth a comment or a different sentinel.
- The `crl_auto_delete_expired_revoked` config key has no UI in this PR (the description says an admin can enable it via the settings UI — that part doesn't exist yet).
- `import uuid as _uuid` inside the function (project style is top-level imports); the three-variant serial lookups duplicate what `utils/serial_format.serial_variants` already provides.
- No CHANGELOG entry.

## Verified OK

- Migration 078: correct numbering, follows the dual-backend pattern, idempotent.
- `RevokedSerial` duck-types cleanly through the CRL/delta builders (`serial_number`, `revoked_at`, `revoke_reason`, `getattr`-guarded `invalidity_at`).
- `superseded` is covered by the OCSP reason map; excluding expired certs from the CRL is fine per RFC 5280; the `caref IS NULL` guard for externally-issued certs (second commit) is correct.

Happy to re-review once the blocking items are addressed — the underlying feature is genuinely worth landing.

NeySlim:
Thanks for this substantial piece of work — the deletion guard, the persistent `revoked_serials` table and the OCSP fallback are exactly right from a PKI-correctness standpoint, and keeping revocations visible in CRL/OCSP after row deletion fixes a real gap. The `crl_auto_delete_expired_revoked` default-off choice is the right call too.

I have one significant concern with the renewal rework: **revoke + delete + create-new-row changes the certificate's `id` and `refid` on every renewal**, which is a breaking change with several side effects:

- **API consumers**: any integration that stores a cert ID (fetch/download/renew loops) gets a 404 after the first renewal. The response and the `cert_renewed` webhook do carry the new ID, but existing automations will break.
- **`approval_requests.certificate_id`**: `delete_certificate` deletes these rows, so every renewal destroys the approval trail of the original issuance — a compliance/traceability regression.
- **`acme_orders.certificate_id`** (FK without `ondelete`): on PostgreSQL, deleting a cert referenced by an ACME order raises an FK violation; the delete fails with only a `logger.warning`, leaving *both* the old revoked row and the new one in place.
- **`discovered_certificates.ucm_certificate_id`** (`SET NULL`): the discovery↔cert link is lost until the next scan re-matches.
- **MSCA FKs** (`csr_id`/`cert_id`) can end up dangling the same way.
- **Audit history** fragments: entries for the old cert point at a dead ID, so you can no longer follow a cert's renewal history on one object.

(The new serial itself is fine — the previous in-place code already generated a new serial on renewal, as it must.)

**Suggestion**: the `revoked_serials` table you're introducing actually makes the delete+recreate unnecessary. The renewal could keep the in-place update and simply insert the `RevokedSerial` record for the *old* serial before overwriting:

1. `RevokedSerial(caref, old_serial, reason='superseded', valid_to=old_valid_to, certificate_id=cert.id)`
2. update `crt`/`prv`/`serial_number`/`valid_from`/`valid_to` in place, `revoked=False`

This composes cleanly with the queries in this PR: the CRL "orphaned revocations" query already includes `RevokedSerial` rows whose `certificate_id` points at a non-revoked cert, and the OCSP path finds no `certificates` row for the old serial and falls back to `RevokedSerial` → `revoked`. Full PKI correctness, zero breaking change, FKs and approval/audit history intact — and the deletion guard + persistence + purge logic in this PR are untouched.

If you'd rather keep one row per generation (which has its own merits for auditability), the FK references above would need to be re-parented onto the new row instead of deleted/orphaned, and the ID change called out as a breaking change in the release notes.

Happy to discuss — the rest of the PR looks solid and CI is green across both backends.

gb-123-git:
@NeySlim 

Thanks for your suggestions.

> The "temporary: added debug logging" commit leaks exception details into API responses

This was a temporary measure which by chance got committed to the online repo instead of just being local. I removed it in the next commit only. Please do check if it still remains by mistake.

> I have one significant concern with the renewal rework: revoke + delete + create-new-row changes the certificate's id and refid on every renewal, which is a breaking change with several side effects:

Your concerns for the breaking changes are very valid.
At the outset, I think we should go with your 1st suggestion of not deleting the certificate and just doing in-place update with old certificate number being added to revoked_serials table. Simple solution.

However, lets think on how we can add auditability of the second option while doing the 1st option.
I'll try and work on it tomorrow.

Let me know if you can also think of something.

gb-123-git:
@NeySlim 

Don't merge as yet. I want to re-check this tomorrow.

gb-123-git:
@NeySlim - I think this is ready.
Please do check this and let me know your views.
Thanks a lot !

NeySlim:
Merged — thank you for the persistence on this one, the final result is a solid piece of PKI work.

What landed (2.214-dev):
- The in-place renewal rework (`id`/`refid`/`created_at` preserved) resolves the breaking-change concern completely, and the single shared implementation across manual/bulk/auto renewal is a real maintainability win
- The atomic revoke/unhold transactions and the revoked-certificate renewal guard are exactly right
- Nice catch on the notification-settings round-trip bug in the last commit

Two small touch-ups applied at integration:
- The i18n keys for the new auto-purge toggle were missing from the locale files — added for all 9 languages
- The CA key loader error detail is now logged server-side instead of being returned in the API response (house rule: no internals in 5xx bodies)

On the CHANGELOG: I rewrote the [Unreleased] entries in the house style, crediting you & PMGA Tech LLP. One ask for future PRs: please leave already-released entries untouched — I'd rather handle attribution changes to published releases case by case.

Thanks again — this closes a real PKI-correctness gap.

PR #304, opened 2026-08-28
fix(de.json): the createCA string is french but should be german

gladiac:
- Replaced the "Créer CA" text in createCA with "CA erstellen".

### DIFF
--- a/frontend/src/i18n/locales/de.json
+++ b/frontend/src/i18n/locales/de.json
@@ -112,7 +112,7 @@
     "clientInfo": "Client-Informationen",
     "exportedFormat": "Exportiert als {{format}}",
     "viewAll": "Alle anzeigen",
-    "createCA": "Créer CA",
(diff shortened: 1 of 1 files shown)


NeySlim:
Thanks @gladiac, good catch. Merged into `dev` with your commit as is (releases flow dev → test → main, which is why the PR is being retargeted to `dev`). The same French label had slipped into the Spanish, Italian, Japanese, Portuguese and Ukrainian locales as well; those are fixed in a follow-up commit on `dev`. Ships in the next release (2.216), credited in the changelog.

NeySlim:
Closing by hand: GitHub refuses to move the base to `dev` now that `dev` already contains your commit ("no new commits between base and head"), so it cannot flag this PR as merged on its own. It is merged nonetheless: `dev` @ 508d20ee3 carries your commit 71518dcb8 with your authorship, and it reaches `main` with 2.216.

Issue #307, opened 2026-08-30
ACME Proxy returns 403 Unauthorized when multiple accounts request the same domain


Seems2Legit:
First of all, thank you very much for your support on my previous issues! 
I adjusted my setup according to your recommendations and it works great, except for one specific edge case regarding multiple accounts.

I am running a setup with multiple Traefik instances, which results in the same domains being requested multiple times by different ACME accounts.
When two different accounts try to request a certificate for the same domain, UCM blocks the authorization for the second account, resulting in a 403 error.

"Require External Account Binding" is disabled for testing.

**Logs UCM**
```
*IP* - - [30/Aug/2026:13:37:42 +0000] "POST /acme/proxy/new-order HTTP/1.1" 201 1336 "-" "containous-traefik/3.7.12 xenolf-acme/5.3.1 (release; linux; amd64)"
2026-08-30 13:37:43 [services.acme.acme_proxy_service] WARNING ACME proxy: refused authorization access on order 1 (requested by foreign account *ACCOUNT*)
*IP* - - [30/Aug/2026:13:37:43 +0000] "POST /acme/proxy/authz/*TOKEN* HTTP/1.1" 403 816 "-" "containous-traefik/3.7.12 xenolf-acme/5.3.1 (release; linux; amd64)"
```

**Logs Traefik**
```
2026-08-30T13:37:43Z ERR Unable to obtain ACME certificate for domains error="unable to generate a certificate for the domains [request.example.com]: acme: error: 403 :: POST :: https://host01.example.com:8443/acme/proxy/authz/*TOKEN* :: urn:ietf:params:acme:error:unauthorized :: Authorization does not belong to this account" ACME CA=https://host01.example.com:8443/acme/proxy/directory acmeCA=https://host01.example.com:8443/acme/proxy/directory domains=["request.example.com"] providerName=ucm-proxy.acme routerName=*ROUTER* rule="*RULE*"
```

**Environment**
UCM Version: v2.216
Traefik Version: 3.7.12

Hemsby:
Thanks for the detailed logs, they pin it exactly. The `refused authorization access on order 1 (requested by foreign account ...)` line is the tell.

The proxy signs every downstream order with a single upstream account, so Let's Encrypt reuses one authorization across every downstream account that requests the same domain. When your second Traefik account fetches that shared authz, the proxy was resolving it to the first account's local order, and the account-ownership check added in #260 then rejected it with the 403 you saw.

[...]

PR #310, opened 2026-08-30
fix(acme-proxy): resolve shared upstream authz to the requesting account (#307)

Hemsby:
## Problem (#307)

With multiple ACME clients that each hold their own account (the reporter runs several Traefik instances), a certificate request for a domain a second account already ordered fails:

```
WARNING ACME proxy: refused authorization access on order 1 (requested by foreign account ...)
"POST /acme/proxy/authz/... HTTP/1.1" 403
urn:ietf:params:acme:error:unauthorized :: Authorization does not belong to this account
```

The proxy signs every downstream order with a single upstream account, so Let's Encrypt reuses one authorization (and its challenge) across every downstream account that orders the same domain. `get_authz`, `respond_challenge` and `_resolve_challenge_order` resolved that upstream URL to the *first* matching local order, and the #260 ownership check then rejected the second account.

## Fix

**Requester-aware resolution.** Among the local orders that recorded the same upstream authz URL, resolution now picks the one owned by the requesting account. `_select_owning_order` does this in two passes (direct owner-field match first, the half-populated-binding reconciliation only if that misses) to keep the extra `AcmeAccount` lookup off the common path. With no owner match it returns the lowest-id candidate unchanged, so `_verify_order_ownership` still issues the denial and #260 is preserved (covered by the new `test_third_account_still_denied`).

**Automation fires once.** The shared challenge's dns-01 automation must not run per-account. The `initiated` flag is now read across every sibling order that shares the upstream authz, not just the order being served. `respond_challenge`'s fallback trigger now persists that flag before starting its thread, mirroring `get_authz`.

In the tight race where both accounts fetch the authz before either commits the flag, both automation threads derive the same TXT value from the shared upstream account thumbprint and the same token, so they write an identical record. Cleanup after either order finalizes is safe because the shared authorization is already valid by then.

**Cache scoping.** `_challenge_order_cache` is keyed per requester identity so a cached `(order id, domain)` is not handed to another account.

## Testing

- `_select_owning_order`, `_challenge_cache_key` and `_challenge_automation_started` verified standalone.
- Three cases added to `backend/tests/test_acme_proxy_ownership.py`: each account resolves to its own order, a third account is still denied, a sibling's initiated challenge blocks a second automation.
- The full pytest suite does not run on my Windows box: `app.py`'s session-directory permission guard rejects the test app factory there (pre-existing, reproduced on a clean checkout). The new tests are written to run in your environment.
- Not exercised live: reproducing two accounts requesting the same domain through a real DNS provider is a race and was not run end to end.

No `CHANGELOG.md` changes.

### DIFF
--- a/backend/services/acme/acme_proxy_service.py
+++ b/backend/services/acme/acme_proxy_service.py
@@ -81,7 +81,7 @@ _LINK_REL_UP_RE = re.compile(
 _cache_lock = threading.Lock()
 _directory_cache = {}        # upstream directory URL -> (stored_at, directory)
 _finalize_url_cache = {}     # upstream order URL     -> (stored_at, finalize URL)
-_challenge_order_cache = {}  # upstream challenge URL -> (stored_at, (order id, domain))
+_challenge_order_cache = {}  # per-requester challenge key -> (stored_at, (order id, domain))
 _nonce_pool = {}             # upstream directory URL -> [(stored_at, nonce), ...]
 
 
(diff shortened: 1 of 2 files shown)


Hemsby:
Tested on Linux (Python 3.13.5, pytest 9.0.3):

- `backend/tests/test_acme_proxy_ownership.py`: 24/24 pass.
- Regression check: reverting only `acme_proxy_service.py` to the pre-fix version, keeping the new tests, fails `test_each_account_resolves_to_its_own_order` (the #307 403) and `test_sibling_initiated_challenge_blocks_second_automation` (duplicate dns-01 automation); `test_third_account_still_denied` passes both ways, so #260 is not regressed.
- Wider ACME proxy / ARI / client tests: 178 passed, 1 skipped.
- Full backend suite: 3868 passed, 8 skipped. Two unrelated failures in `test_deploy_hooks.py` from `paramiko` missing in the test venv.

Not exercised: a live two-account run against a real DNS provider.

Issue #312, opened 2026-08-30
TSA: tsa_require_dedicated_cert has no way to actually supply a dedicated signer


Hemsby:
### The gap

The `tsa_require_dedicated_cert` toggle (Settings > TSA) can be turned on, but there is no way to satisfy it. The `/tsa` endpoint always signs with the configured CA's own certificate (`tsa_protocol.py` loads `ca.crt` / `ca.prv` from the CA at `tsa_ca_refid`), and there is no setting anywhere to point it at a dedicated end-entity certificate. So enabling the toggle just makes `TSAService` raise and every timestamp request returns 503, with no path forward for the operator.

This came up via #309 / #311. #311 fixes the immediate bug (a constrained sub-CA certificate no longer 503s), but the underlying story is that RFC 3161 §2.3 wants a dedicated signer with a critical, exclusive `timeStamping` EKU, and UCM can't currently produce or use one.

### Proposed approach

Add `tsa_signer_cert_refid`:

- **When set:** load that certificate + key as the timestamp signer instead of the CA certificate. Validate it is not expired / revoked at signing time. Resolve by refid so a renewal (which re-keys) is picked up automatically.
- **When unset:** current CA-certificate behavior, unchanged.
- Embed the signer's issuer chain in the token (`_build_signed_response` currently embeds only the leaf), so strict verifiers can build a path.
- Config API (`api/v2/tsa.py`) returns signer subject / serial / notAfter / chain status; PATCH sets or clears it.
- TSA page gets a "Signing certificate" section (CA certificate vs dedicated), shows expiry and chain validity, and gates the `require_dedicated_cert` toggle on a valid dedicated signer being present.

Two ways to get the certificate itself:

1. Let the operator select any already-issued certificate that carries the `timeStamping` EKU (`validate_certificate` already checks this). Small.
2. A one-click "generate TSA signing certificate" that issues an end-entity cert from the TSA CA with a critical, exclusive `timeStamping` EKU, key stored encrypted in UCM. The sign-CSR path already supports the EKU (per the 2.201 delegated-responder / TSA work). Larger, pulls in issuance and renewal.

### Lifecycle notes

- An end-entity signer expires, unlike the CA cert. A silently expired signer means 503s return. Near-expiry should be surfaced on the dashboard, and auto-renew is worth considering.
- A revoked signer should stop being used.
- Already-issued timestamps are self-contained and unaffected.

[...]

PR #313, opened 2026-08-31
feat(tsa): dedicated end-entity signing certificate (#312)

Hemsby:
Closes the gap in #312: `tsa_require_dedicated_cert` could be turned on but nothing could satisfy it, because `/tsa` always signed with the configured CA's own certificate.

This is your "option 1 only" scope from the issue thread.

## What it does

Adds a `tsa_signer_cert_refid` SystemConfig key (no schema change). When set, `/tsa` loads that certificate and its key as the timestamp signer instead of the CA certificate.

- **Resolution by refid at signing time**, so an in-place renewal is picked up automatically.
- **Re-checked on every request**: an expired, revoked, or undecryptable signer fails the request with 503 and an explicit log line. It never falls back to the CA certificate.
- **Issuer chain embedded**: the token carries the signer leaf plus its issuing CA chain when the client sends `certReq`. `_build_signed_response` previously embedded only the leaf.
- **ESSCertIDv2 / SigningCertificate** signed attribute follows the selected signer (it already derived from `self.tsa_cert`, which is now the dedicated signer).
- **Config API**: `GET /api/v2/tsa/config` returns `signer_cert_refid` and a resolved `signer` object (subject, serial, notAfter, chain status, usability). `PATCH` validates the selection. New `GET /api/v2/tsa/signer-candidates` lists eligible certificates.
- **TSA page**: a "Signing certificate" section with the picker and a live status box. The `require_dedicated_cert` toggle stays inoperable until a usable dedicated signer is saved, and it is cleared automatically if the signer is removed, so the UI cannot leave strict mode set with no signer.

With `tsa_signer_cert_refid` unset, the emitted token is byte-for-byte what it is after #311 (`chain_certs` defaults to empty, nothing else in the path changes). `TestDedicatedSignerToken::test_ca_signer_path_embeds_exactly_one_certificate` is the regression guard.

## Constraints from the issue thread, as implemented

- The candidate picker only offers certificates whose private key UCM holds and can decrypt: `private_key_location == 'stored'`, a key is present, and `source` is not one of `acme` / `letsencrypt` / `scep` / `est` / `msca`. Certificates carry no `hsm_key_id` in the model (only CAs do), so there is nothing HSM-shaped to exclude at the certificate level.
- An expired or revoked signer fails hard (503 + explicit log), never a silent CA fallback. The signer is an ordinary UCM certificate, so the existing expiry alerts and in-place auto-renewal are the safety net.
- CHANGELOG entry under `[Unreleased]` this time.

1. **The signer is not required to chain to the TSA CA.** It can be issued by any CA. The config API reports chain status (`chain_to_root`) rather than enforcing it. This seemed right given the signer stands alone, but say the word if you want it constrained to the TSA CA.
2. **`signer-candidates` does not attempt to decrypt every candidate key**, only checks that a stored key is present and the EKU matches. The actual decrypt attempt happens at `PATCH` time on the one selected certificate. Decrypting every key on each settings load felt wasteful. The `PATCH` gate is the real enforcement.
3. **The backend stores `require_dedicated_cert` as sent.** The gating lives in the UI per your scoping. I did not add a backend refusal of `require_dedicated_cert: true` with no signer. Easy to add if you would rather it be enforced server side.
4. **English strings only.** The other 8 locales fall back to `en` until you do a translation pass.
5. Building against your dev tip, I found that all issuance paths need the operator to hand-add the timeStamping EKU (none of my lab certificates had it). That reinforces the case for option 2 (one-click issuance) as the follow-up you described.

## Testing

- `backend/tests/test_tsa_service.py`: 26 pass on a Linux box, including new unit tests for chain embedding and the byte-for-byte regression guard, and HTTP-level tests for the candidate list, PATCH validation, a full `/tsa` round trip with a dedicated signer, and the fail-hard-no-fallback path.
- Real `openssl ts -verify` accepts a dedicated-signer token with the embedded chain.
- Frontend: new and updated vitest coverage for the picker, the status box states, and the toggle gating. Full page suite green, production build clean.
- Manually exercised end to end on a lab instance: issue a timeStamping cert, select it, save, verify the token, clear it, confirm strict mode releases.

Follow-ups you called out and I have left out: one-click TSA signer issuance, and dashboard surfacing of a near-expiry signer.


NeySlim:
Merged, with the 8 missing locales translated at integration (the i18n structural check is part of the pre-commit gate here, which is why your build was fine but the hook would have refused it). Ships in 2.217.

Answers to your three questions:

1. **Signer not constrained to the TSA CA: agreed.** A signer issued by any CA is legitimate; reporting `chain_to_root` instead of enforcing it is the right call.
2. **No decrypt-all in the candidates list: agreed.** The PATCH-time decrypt on the selected certificate is the real gate; probing every key on settings load would be waste.
3. **UI-side gating of `require_dedicated_cert` is fine as shipped.** The strict mode already fails hard at signing time server-side, so a stored-but-unsatisfied toggle cannot mint CA-signed tokens; a server-side refusal at PATCH would be belt-and-braces, not required.

Quality note: the byte-for-byte regression guard and the certReq-off test are exactly the kind of coverage that makes these reviews fast. The one-click TSA signer issuance and the near-expiry surfacing stay open as follow-ups (#312 remains the tracker).

PR #314, opened 2026-08-31
feat(tsa): one-click issuance of a dedicated TSA signing certificate (#312)

Hemsby:
The "option 2" follow-up from #312. #313 made `tsa_signer_cert_refid` satisfiable, but as you noted in the issue thread, a certificate issued through the regular API cannot satisfy a strict RFC 3161 verifier: `cert_create.py` always emits the EKU extension non-critical and merges the base profile with `extra_ekus`, so `timeStamping` is never critical or exclusive. The operator was left hand-crafting the certificate outside UCM.

## What it does

Adds `POST /api/v2/tsa/signer-certificate` and a "Generate a signing certificate" action in Settings > TSA > Signing certificate. It issues an end-entity certificate with the extensions built directly rather than through the generic issuance profile:

| Extension | Value |
|---|---|
| BasicConstraints | CA:FALSE, critical |
| KeyUsage | `digitalSignature` only, critical |
| ExtendedKeyUsage | `timeStamping` only, **critical** |

plus SKI, AKI, and the issuing CA's CDP / AIA / CPS when enabled.

- The private key is stored encrypted at rest, the same path `renewal.py` uses.
- The certificate is issued and committed first. When no usable dedicated signer is configured yet, it is then selected as `tsa_signer_cert_refid` in a second write; if that write fails the certificate stays issued but unselected and a log line records it. An already-healthy signer is never swapped without an explicit `select: true`.
- Defaults: the configured TSA CA, `RSA 3072`, 397-day validity clamped to the issuing CA's own expiry rather than refused (the operator did not pick the validity in the one-click flow).
- `cert_type` is `timestamping`, `source` is `manual`.
- `describe_configured_signer()` and `/signer-candidates` now report `eku_critical_exclusive`, surfaced on the TSA page as an "RFC 3161 strict" badge on candidates and a line in the signer status box. This is the visible signal you described.

With `tsa_signer_cert_refid` unset, nothing changes.

## Constraints from the thread, as implemented

- The generated certificate has a critical, exclusive `timeStamping` EKU, KU restricted to `digitalSignature` (critical), and BasicConstraints CA:FALSE, built explicitly rather than via `cert_create.py`, exactly as you specified.
- Auth: the endpoint requires both `write:settings` and `write:certificates` (`require_auth` treats the list as AND). Minting a certificate from a settings page is a privilege widening, so both scopes felt right. The Generate button is gated on both.
- The certificate is an ordinary UCM certificate. In-place renewal (`renewal.py` copies extensions with their criticality) keeps the critical exclusive EKU, and the standard expiry alerts apply.
- CHANGELOG entry under `[Unreleased]`.
- English strings only; the other 8 locales fall back to `en`.

1. **Auto-renewal coverage.** `AutoRenewalService` filters on `source in ['scep','acme','est']` by default, so a `source='manual'` signer is not auto-renewed unless the operator adds `manual` to `auto_renewal_sources`. Expiry alerts still fire. I kept `source='manual'` to stay inside the existing certificate model rather than introduce a dedicated `source='tsa'` and touch the auto-renew defaults, but say the word if you would rather it be covered out of the box.
2. **Two writes, not one.** The helper commits the certificate, then the route does a second commit for the selection. A failed selection leaves an issued, unselected certificate (harmless, logged). I can thread a `commit=False` flag through the helper if you want a single transaction.
3. **Key default `RSA 3072`**, with `RSA 2048/4096` and `EC P-256/P-384` selectable in the modal. Chosen for the broadest timestamp-verifier compatibility.

## Testing

- `backend/tests/test_tsa_service.py`: new `TestTsaSignerOneClickIssuance` covering the strict extensions, the candidate report, auto-select vs no-swap of a healthy signer, validity clamping against a short-lived CA, a full `/tsa` round trip signing with the generated signer, and the permission gate.
- `frontend/src/pages/__tests__/TSAPage.signerIssue.test.jsx`: the modal flow and the strict/loose EKU rendering. Full TSA page suite green, production build clean.
- Backend suite runs in CI (my dev box is Windows, where `app.py` cannot import).
- Exercised end to end on a lab instance against the merged #313 code: generate a signer with defaults, confirm `usable: true` and `eku_critical_exclusive: true`, `openssl ts -verify` accepts the token (this is what failed before), enable strict mode and confirm `/tsa` still grants, revoke the signer and confirm `/tsa` returns 503 with an explicit log line and no CA fallback.

Follow-up still open on #312: dashboard surfacing of a near-expiry signer.


NeySlim (changes requested):
Thanks, this is exactly the shape discussed in #312: the crypto content is right where the generic issue path cannot be. Verified on the diff: `BasicConstraints` CA:FALSE critical, `KeyUsage` digitalSignature only critical, and the `timeStamping` EKU critical and exclusive, which is what a strict RFC 3161 verifier wants. The auto-select policy (never swap a healthy signer without an explicit request, selection failure leaves the certificate issued and logged) is also the right call, and the tests cover it.

Requesting changes on four points before merge:

1. **i18n is EN-only (P2)**: the 15 new keys (the `tsa.signerGenerate*` block plus `certificates.types.timestamping`) were added to `en.json` only. All 9 locale files must stay in sync (`node scripts/check-i18n-sync.js` fails on this branch). Same note as on your previous PRs: the other 8 locales need real translations, not copies.

2. **Validity 0 is silently swallowed (P2)**: in `TSAPage.jsx`, `validity_days: Number(genForm.validity_days) || undefined` maps an explicit `0` to `undefined`, so the backend applies the 397-day default instead of returning its own "validity_days must be positive" 400. The user asked for 0 and gets a 397-day certificate with a success toast. Send the raw value when the field is non-empty and let the backend validate it.

3. **API input validation (P2)**: two cases on `POST /api/v2/tsa/signer-certificate`:
   - `want_select = bool(data['select'])` treats any non-empty string as true, so `{"select": "false"}` swaps a healthy signer. Require a real boolean (`isinstance(data['select'], bool)`, else 400).
   - a non-string `cn` (e.g. `{"cn": 123}`) reaches `(cn or DEFAULT_CN).strip()` and 500s. Same guard: reject non-string `cn` with a 400. We just fixed the identical pattern on the ACME account PATCH, so let's not reintroduce it here.

4. **Auto-renewal blind spot, needs to be explicit (decision)**: the certificate is issued with `source='manual'`, and the auto-renewal scheduler only covers `scep, acme, est` by default, so the dedicated signer of a critical endpoint will not renew itself on a default install. The CHANGELOG says so, but nobody reads the CHANGELOG in the generate modal. At minimum the modal (and/or the status box) should state it and point to the renewal settings; alternatively we discuss covering the configured TSA signer in auto-renewal regardless of source. Tell me which way you want to take it, the first option is enough for this PR.

Points 1 to 3 are mechanical; happy to merge once they are in and the answer to 4 is at least a UI hint.

Hemsby:
Thanks. All four addressed in 077114996.

**1 (i18n).** The 15 keys (`tsa.signerGenerate*` / `signerStrictBadge` / `signerEku*` plus `certificates.sourceTypes.timestamping`) are now translated in all 8 other locales, not copied. `check:i18n` is green. I also swapped the two em dashes in the English `signerEkuStrict` / `signerEkuLoose` strings for semicolons while I was in there.

**2 (validity 0).** `TSAPage.jsx` now sends `validity_days` whenever the field is non-empty (`0` included) and lets the backend answer. An explicit `0` gets the "validity_days must be positive" 400 and an error toast instead of a 397-day certificate with a success toast. Covered by `test_explicit_zero_validity_is_rejected_not_defaulted` and two frontend cases.

**3 (input validation).** On `POST /api/v2/tsa/signer-certificate`, before anything reaches the issuer:

- `select` must be a real `bool` (`isinstance(..., bool)`), so `{\"select\": \"false\"}` is a 400 and never swaps a healthy signer.
- a non-string `cn` is a 400, not a 500 at `.strip()`. `tsa_signer_cert.py` carries the same `cn` guard as defence in depth.

Tests: `test_non_boolean_select_is_rejected`, `test_non_string_cn_is_rejected_with_400_not_500`.

**4 (auto-renewal blind spot).** Took the UI-hint option. The Generate modal now spells out that the certificate is issued with source `manual` and that automatic renewal only covers it if `manual` is added to the renewal sources in Settings > Automatic Renewal. Covered by `test shows the auto-renewal caveat in the Generate modal`. Happy to also discuss covering the configured TSA signer in auto-renewal regardless of source as a follow-up if you'd prefer that longer term.

Backend `test_tsa_service.py` green on Linux/py3.13 (37, up from 34); frontend `TSAPage.signerIssue` green (7, up from 4).

Heads-up: this and #315 now both touch `tsa_service.py` (#315 extracts `tsa_is_enabled()` to share the enable rule with `tsa_protocol.py`), so whichever lands second needs a small rebase.

NeySlim (approved):
Re-verified on the new diff, all four points are properly addressed:

1. The 15 keys are present in all 9 locales with real translations (spot-checked de/es/fr/it), and the renewal note reads well everywhere.
2. An explicit 0 now reaches the backend and comes back as the "must be positive" 400 with an error toast; the field being type="number" closes the NaN path.
3. `select` requires a real boolean and a non-string `cn` is a 400 in the route with the same guard in the service; both tests are collected and green.
4. The modal note on `source "manual"` and renewal sources is exactly the explicitness I asked for.

CI is green across the matrix. Approving.

PR #315, opened 2026-08-31
feat(dashboard): surface a near-expiry TSA signer (#312)

Hemsby:
The follow-up you deferred on #312: dashboard surfacing of a near-expiry dedicated signer, "aligned with the existing expiry-alert plumbing."

Independent of #314. It only needs the signer concept from #313, already in `dev`.

## Why

The dedicated signer from #313 is a single point of failure for `/tsa`: it returns 503 the moment the signer expires, is revoked, or its key stops decrypting, with no CA fallback. The CA certificate the TSA historically signed with outlives everything; an end-entity signer lapses on its own schedule, and nothing on the dashboard warned about it.

## What it does

`GET /api/v2/dashboard/system-status` gains a `tsa` entry, from `describe_configured_signer()` plus `tsa_enabled`:

| condition | status | message |
|---|---|---|
| disabled | `offline` | "Disabled" |
| enabled, no dedicated signer | `online` | "Signing with CA certificate" |
| enabled, signer within 30 days of expiry | `warning` | "Signer expires in N day(s)" |
| enabled, signer expired / revoked / undecryptable | `offline` | "Signer unusable: ..." |
| enabled, signer healthy | `online` | "Dedicated signer" |

Wrapped like the sibling checks; any failure falls back to `online` / "Status unknown" so a resolver bug never blanks the widget. The 30-day threshold matches the auto-renewal window already used in the same function.

`GET /api/v2/dashboard/expiring-certs` gains `is_tsa_signer` per row.

Frontend:
- `ServiceBadge` now renders three states (online, amber warning, grey) and shows its status message on hover. This also fixes EST, SMTP and Webhooks, which already emit `warning` but rendered identically to offline.
- A `TSA` badge is added to both System Health layouts.
- The Next Expirations widget tags the configured signer with a "TSA signer" chip so its expiry reads as infrastructure rather than one certificate among many.

The signer is an ordinary certificate, so the existing email expiry alerts already cover it. This is the at-a-glance view.

1. **CA-certificate signing shows `online`, not `warning`.** `tsa_service` already logs a warning on every CA-cert sign; a permanent amber dashboard badge for the historical default would nag deployments that run it deliberately. The message names the state. Flip it to `warning` if you want it louder.
2. **`is_tsa_signer` on the expiring-certs payload** is resolved by comparing each cert's `refid` to `tsa_signer_cert_refid`, one extra `SystemConfig` read per call. It does not depend on #314's `cert_type='timestamping'`, so it works for a signer selected before #314 too.

## Testing

- `backend/tests/test_dashboard.py`: `TestTsaSystemStatus` covers the five states (disabled, CA-certificate, near-expiry warning, healthy, revoked); `test_expiring_certs_flags_the_configured_tsa_signer` covers the row flag. 35 pass on Linux (Python 3.13).
- No frontend test: `DashboardPage` has no existing test and a heavy context-mock surface. The change is a `ServiceBadge` instance, a conditional chip, and a three-way class map.
- Exercised on a lab instance: with a healthy signer the badge reads `online` "Dedicated signer"; issuing a 20-day signer flips it to `warning` "Signer expires in 19 day(s)" and the Next Expirations widget shows the "TSA signer" chip; revoking it flips the badge to `offline` "Signer unusable: ... is revoked".

English strings only; the other 8 locales fall back to `en`.


NeySlim (changes requested):
The idea is right (a dedicated signer is a single point of failure and deserves dashboard visibility), the ServiceBadge `warning` state and the hover message are nice additions, and the backend tests are real. But the status logic disagrees with what `/tsa` actually does, in both directions. I reproduced both cases:

1. **P1: status logic contradicts the protocol endpoint** (`backend/api/v2/dashboard.py`):
   - TSA enabled with no dedicated signer AND no TSA CA configured: the widget reports **online** ("Signing with CA certificate"), but `/tsa` refuses to sign without an admin-designated CA and returns 503. The `configured: False` branch must also check `tsa_ca_refid` before claiming the CA path works.
   - Pre-2.200 grandfathered install (TSA CA configured, `tsa_enabled` row never written): the protocol endpoint explicitly treats a **missing** row as enabled (see the grandfathering comment in `tsa_protocol.py`), so `/tsa` works, but the widget's `str(None).lower() != 'true'` shows **offline / Disabled**. The widget must mirror the protocol's rule: only an existing row with a non-true value means disabled.

   Both cases are the exact situations this PR exists to surface, so they need tests too (no TSA CA + enabled, and missing `tsa_enabled` row).

2. **P2: a resolution error renders green**: the `except Exception` fallback reports `online / Status unknown`. An exception in the status check is precisely the situation where the dashboard must not claim the service is healthy. Make it a neutral or `warning` state, never `online`.

3. **P2: raw internal details reach the widget**: `Signer unusable: {signer.get('error')}` forwards `describe_configured_signer()` messages verbatim, and some of them embed the underlying exception text (parse and key-loading failures). The endpoint is plain `@require_auth()`, so any authenticated user sees it. Project rule is generic messages to the client and details in the logs: map the error to a coarse reason (expired / revoked / key unavailable / invalid) and log the rest.

4. **P2: i18n and frontend tests**: `dashboard.tsaSigner` exists in `en.json` only (the other 8 locales need it), and nothing covers the new UI: no test for the `warning` badge state, the hover message, or the TSA signer chip on the expirations widget. The repo now has page-level component tests to crib from.

Happy to re-review quickly, the fix for point 1 is small and the rest is mechanical.

Hemsby:
Thanks. All four addressed in 42b320b55.

**1 (status parity).** I pulled the two rules the widget was guessing at into `tsa_service.py` so it and `tsa_protocol.py` share them:

- `tsa_is_enabled()` mirrors the grandfathering rule (missing `tsa_enabled` row = enabled); `tsa_protocol.py` now calls it too.
- `tsa_ca_certificate_path_ready()` mirrors the CA-path gates (`tsa_ca_refid` set, CA present with cert and key, not offline).

So enabled + no dedicated signer + no usable TSA CA now reports offline ("No signing certificate configured"), and a grandfathered install with a configured CA reports online. Both new cases have tests (`test_offline_when_enabled_but_no_tsa_ca_configured`, `test_grandfathered_missing_enabled_row_mirrors_protocol`), and the existing CA-path tests now seed a real CA.

That does mean `tsa_protocol.py` shows up in this PR. The change there is a straight swap to `tsa_is_enabled()` with no behavior change; `test_protocol_signing_guards.py::TestTsaConfigGuard` (all five cases) still passes. The alternative was writing the rule a third time.

**2 (error renders green).** The `except` fallback is now `warning` / "Status unavailable", never `online`. Test: `test_resolution_error_is_not_reported_as_online`.

**3 (raw detail leak).** `TSAConfigurationError` now carries a coarse `reason` (`expired` / `revoked` / `key_unavailable` / `invalid`). The widget maps that to a fixed string and logs the full message; the `read:settings` config API still gets the detailed `error`. Test: `test_offline_message_does_not_leak_internal_detail` drops the private key and asserts no refid and no traceback in the client message.

**4 (i18n + frontend tests).** `dashboard.tsaSigner` is now in all 9 locales (`check:i18n` green). New `DashboardPage.tsaSigner.test.jsx` covers the warning badge state, the `title` hover message, and the signer chip on the expirations widget (present for `is_tsa_signer`, absent otherwise).

Backend suite green on Linux/py3.13 (79 across `test_dashboard.py`, `test_tsa_service.py`, `test_protocol_signing_guards.py`); frontend suite green including the 3 new tests.

Heads-up: this now touches `tsa_service.py`, which #314 also edits (`eku_critical_exclusive`), so whichever lands second needs a small rebase.

NeySlim (approved):
Re-verified on the new diff, all four points are properly addressed, and pulling the two rules into `tsa_service.py` helpers is the right call:

1. `tsa_is_enabled()` is an exact behavioral match for the old inline check (missing row = enabled), the `tsa_protocol.py` swap uses its return value correctly, and `tsa_ca_certificate_path_ready()` mirrors the CA-path gates. Both review cases have tests, and the autouse clear fixture around each test keeps the shared-DB state clean (including `tsa_ca_refid`).
2. A resolver exception now reports `warning` / "Status unavailable", never online.
3. The coarse `reason` mapping with the fixed message table keeps the raw text in the logs and out of the widget; the detailed message staying on the `read:settings` config API is the right split.
4. `dashboard.tsaSigner` is in all 9 locales and the three new page tests cover the warning badge, the offline state and the chip.

CI is green across the matrix. Approving.

Issue #316, opened 2026-09-02
Create CA: no way to set Name Constraints


JoseGoncalves:
## What I'm trying to do

Create an internal root CA restricted to my organisation's domain, so it can only ever issue certificates for names underneath it.

## What happens

In the Create CA wizard, under **Advanced Constraints (RFC 5280)**, the only options are *Path Length Constraint* and *Use CA name in protocol URLs*. There is no field for name constraints, permitted or excluded.

I also tried creating the CA through the REST API instead, but couldn't find any value for the name-constraint parameters that it would accept.

## What I expected

Fields under *Advanced Constraints* for permitted and excluded name constraints (DNS, IP, email), offered the same way Path Length is.

## Why it matters

Name constraints are what stops a private root from being able to issue for names outside your own namespace. Without them, if the CA key is ever compromised it can mint a trusted certificate for *any* domain, on every machine that carries the root.

They also can't be added later — the extension lives inside the signed certificate — so this has to be available at creation time. Editing the CA afterwards wouldn't help.

## Environment

UCM 2.219 (`.deb`), Debian 13, Root CA (self-signed).

Is there another way to set these that I've missed?

Hemsby:
Thanks for the detailed write-up, and you're right on both counts.

The Create CA wizard doesn't expose permitted or excluded name constraints anywhere. The backend does have NameConstraints support already: the storage exists, the extension is built as critical, and issuance and renewal are validated against the CA chain's constraints. There's just no way to set them when the CA is created.

On the REST API, `POST /api/v2/cas` does accept `nameConstraintsPermitted` and `nameConstraintsExcluded`, but I hit the same wall you did. Looking into why, the request validation currently expects plain strings while the code that builds the extension expects `{"type": "dns|ip|email", "value": "..."}` objects, so neither shape gets you a constrained CA. That explains why nothing was accepted.

@NeySlim I'm going to start on a PR for this: fields for permitted and excluded constraints (DNS, IP, email) under Advanced Constraints alongside Path Length, and aligning the API so it accepts and stores them. I'll link it here once it's up.

Hemsby:
PR is up: NeySlim/ultimate-ca-manager#317

[...]

PR #317, opened 2026-09-02
feat(cas): name constraints in the Create CA wizard and API (#316)

Hemsby:
## Problem

Issue #316: no way to set RFC 5280 name constraints when creating a CA.

- The Create CA wizard's Advanced Constraints section only offers Path Length and the protocol-URL toggle.
- `POST /api/v2/cas` accepts `nameConstraintsPermitted` / `nameConstraintsExcluded`, but the request validation added in 0a01de311 requires plain string entries, while `ConstraintsMixin._build_general_subtrees` expects `{type, value}` objects. A dict payload is rejected at validation; a string payload passes validation and then cannot be turned into an extension. No input produces a constrained CA.

Name constraints can't be added after issuance, so this has to work at creation time.

## Changes

**API** (`backend/api/v2/cas/crud.py`)
- `_normalize_name_constraints()` validates each list as `{type, value}` objects (`type` in dns / ip / email, case-insensitive; value trimmed, non-empty, <= 255 chars; list <= 32 entries).
- `ip` values are checked with the same `ipaddress.ip_network()` call the builder makes, so `10.0.0.5/8` is a 400 rather than a 500 during cert creation.
- Malformed entries are rejected, not skipped: a dropped entry would produce a CA with no NameConstraints extension, the opposite of the request.
- Name constraints with `type: external` are refused, since `create_external_ca` can't apply them (the signing CA does).

**Wizard** (`frontend/src/pages/cas/CreateCAModal.jsx`)
- `NameConstraintRows`, modeled on the SAN editor in `IssueCertificateForm`: repeatable rows of type select (DNS / IP range / Email) + value + remove, for permitted and excluded subtrees, in the Advanced Constraints section.
- Hidden for external-CSR CAs, which show a one-line note instead.
- Blank rows are dropped on submit.

**i18n** — 9 new `cas.*` keys with translations in all 9 locales; reuses the existing `nameConstraintsPermitted` / `nameConstraintsExcluded` labels.

**CHANGELOG** — `[Unreleased]` > Added.

## Tests

- `backend/tests/test_ca_name_constraints_api.py` (9): a CA created through `POST /api/v2/cas` carries the critical NameConstraints extension with the expected permitted/excluded subtrees; an unconstrained CA has none; a violating certificate is rejected under the API-created CA; the six 400 cases. The existing enforcement tests hand-build the constrained CA cert, so this path was untested.
- `frontend/src/pages/cas/__tests__/CreateCAModal.nameConstraints.test.jsx` (2): permitted rows serialize as `{type, value}` and blank rows drop; external CA hides the fields and omits constraints.

## Verification

- Backend suite on Linux/py3.13 (isolated DB): 140 passed across the new file plus `test_name_constraints_enforcement`, `test_name_constraints_renewal_grace`, `test_cas`, `test_ca_external_csr`.
- `npm run build`, eslint, vitest clean.
- Deployed to a test box: backend healthy, endpoint reachable.

Reported by @JoseGoncalves.


NeySlim (changes requested):
Thanks, this is exactly the shape described on #316, and the diagnosis is right: verified on `dev` that the request validation required plain strings while `_build_general_subtrees` consumes `{type, value}` objects, so no payload ever produced a constrained CA. Re-ran the branch locally: 140 backend tests green across the new file plus the enforcement, renewal-grace, cas and external-CSR suites, the two wizard tests and eslint green, `check-i18n-sync` passing with real translations in the 8 other locales (thanks for that). Also exercised a few edge cases through `POST /api/v2/cas`: non-ASCII DNS/e-mail values come back 400 with nothing persisted, an IPv6 network is accepted, an intermediate CA carries the critical extension, empty lists emit no extension, and `" DNS "` / `" example.net "` are normalised.

Requesting changes on two points before merge, plus a nit:

1. **Subject emailAddress is never checked against e-mail constraints (P2)**: `validate_name_constraints()` only collects the CN (when it looks like a hostname or contains `@`) and the SANs. RFC 5280 §4.2.1.10 requires rfc822Name constraints to be applied to the `emailAddress` attribute of the subject DN when the certificate has no SAN. Reproduced on the branch: CA created with `nameConstraintsPermitted: [{"type": "email", "value": "example.com"}]`, then

   ```json
   POST /api/v2/certificates
   {"ca_id": <id>, "cn": "John Doe", "email": "user@outside.test", "validity_days": 30}
   ```

   is issued (201). The same address as `san_email` or as the CN is refused with 400 "Names outside CA permitted NameConstraints", as expected. The gap predates this PR (the function is untouched here), but this PR is the first way to create a CA with an e-mail constraint from UCM, so it is the place to close it. Fix in `validate_name_constraints`: add the subject's `EMAIL_ADDRESS` attributes as `RFC822Name` to `names_to_check`; that covers every issuance path (direct, approval, CSR signing, SCEP, renewal, TSA signer) at once. `_cert_name_values` should include the subject emailAddress too so the renewal-at-par grace stays consistent.

2. **DNS and e-mail values are not validated syntactically (P2)**: only `ip` is really checked. All of these are accepted and produce a CA with a critical, irreversible NameConstraints extension: dns `https://example.com`, `foo bar`, `example.com/path`, `example..com`, `-bad.example.com`, `exa_mple.com`; email `not an email`, `https://x`, `@example.com`. Only `*.example.com` and `a@b@c` are refused, and only because cryptography raises. The effect is the worst of both worlds: as a permitted subtree `https://example.com` blocks every DNS name (`host.example.com` refused), so the CA is unusable; as an excluded subtree it matches nothing (`host.example.com` issued), so the constraint is inert. `openssl x509 -text` shows `Permitted: DNS:https://example.com`. Please validate in `_normalize_name_constraints`: dns = LDH labels with an optional leading dot (subtree form), no scheme, path, space, underscore or wildcard; email = either `local@domain` or the domain / `.domain` forms RFC 5280 allows for rfc822Name constraints. Rejecting non-ASCII (or IDNA-encoding it) at the same place gives a clearer 400 than the cryptography message.

3. **CHANGELOG credit (nit)**: the entry ends with `(#316, reported by @JoseGoncalves)`; please add `contributed by @Hemsby` as on #319 and the other entries.

Points 1 and 2 each deserve a test (`test_ca_name_constraints_api.py` is the natural home for 2, `test_name_constraints_enforcement.py` for 1). Happy to merge once they are in.

Hemsby:
Pushed e34df19 with both P2 fixes and the CHANGELOG nit.

1. **Subject emailAddress vs e-mail constraints.** `validate_name_constraints()` now adds the subject DN `EMAIL_ADDRESS` attributes to `names_to_check` as `RFC822Name`, unconditionally. Since every caller (direct, approval, CSR signing, SCEP, renewal, TSA signer) passes the built subject, that closes the gap on all of them at once. `_cert_name_values()` keys the same attribute as `RFC822Name` so renewal at par still graces a DN e-mail that predates the constraint. Your repro (`email` permitted constraint, then a request with `email: user@outside.test` and no SAN) now returns 400; the same address inside the permitted domain issues. Enforcement test plus a renewal-at-par grace test in `test_name_constraints_enforcement.py`.

2. **DNS and e-mail syntax.** `_normalize_name_constraints()` validates `dns` and `email` alongside `ip`: dns must be LDH labels with an optional leading dot, no scheme / path / space / underscore / wildcard; email must be `local@domain`, a bare `domain`, or `.domain`. Non-ASCII is rejected first with an IDNA hint. `https://example.com`, `foo bar`, `example..com`, `-bad.example.com`, `exa_mple.com`, `*.example.com`, `not an email`, `@example.com`, `a@b@c` all 400 now; `example.com`, `.example.com`, `user@example.com`, `example.com`, `.example.com` (email) are accepted. Parametrized cases in `test_ca_name_constraints_api.py`.

   The guard is at the `POST /api/v2/cas` boundary, which is the only way to set constraints on a CA, so I did not touch `_build_general_subtrees`.

3. CHANGELOG `[Unreleased]` entry extended and credited `contributed by @Hemsby`.

Verification: full backend suite on a Linux box / py3.13, 3999 passed / 8 skipped (the new subject-email name type in `names_to_check` trips nothing else). No frontend change in this push.

NeySlim (changes requested):
Re-verified e34df19: the subject emailAddress repro is now a 400, the syntax cases are 400, the CHANGELOG credit is in, and the targeted suites are green here too. Two points remain, one of them territory the e-mail type opens up, plus a related note:

1. **E-mail constraint semantics are wider than RFC 5280 (P2)**: `_name_matches_subtree()` in `validation_helpers.py` treats `example.com` and `.example.com` almost the same for rfc822Name: both accept the exact host and its subdomains (lines 43-48, and the docstring says so). RFC 5280 §4.2.1.10 makes them distinct: `example.com` = mailboxes on that host only, `.example.com` = mailboxes within the domain only, and a full `user@example.com` = that mailbox only. OpenSSL enforces exactly that, so UCM issues certificates that fail chain validation. Reproduced on the branch, then `openssl verify -CAfile ca.pem leaf.pem` on what was issued:
   - permitted `example.com`, SAN `user@sub.example.com`: UCM 201, OpenSSL "permitted subtree violation";
   - permitted `.example.com`, SAN `user@example.com`: UCM 201, OpenSSL "permitted subtree violation".

   The excluded side over-blocks symmetrically (excluded `example.com` refuses `user@sub.example.com`). The code predates this PR, but this PR is what makes e-mail constraints reachable, so let's fix it here: mailbox form = exact match, `.domain` = strict suffix on the domain part, bare `domain` = exact domain match. Tests belong in `test_name_constraints_enforcement.py`.

   Related, since the new syntax check accepts a leading dot for `dns` too: `.example.com` as a dNSName constraint matches `example.com` itself in UCM (line 35) while OpenSSL rejects it (verified: permitted `.example.com`, CN `example.com` issues, `openssl verify` reports a violation). RFC 5280 has no leading-dot form for dNSName. Either reject the leading dot for `dns` at the API boundary, or make the DNS branch subdomain-only; I lean towards rejecting, one less thing to explain in the UI.

2. **Mailbox local part is barely validated (P3)**: `_valid_email_name_constraint()` checks the domain and rejects whitespace, but `a..b@example.com`, `.a@example.com`, `a.@example.com` and `a:@example.com` all land in the extension (201). Same consequence as last round (inert or all-blocking, permanent), but it takes an operator typo, hence P3. Validate the local part as an RFC 5322 dot-atom (no leading, trailing or consecutive dots, atext characters only, at most 64 chars). Quoted local parts can stay unsupported.

Happy to merge once point 1 (and its DNS sibling) is in; 2 is a one-liner in the same file.

Hemsby:
Ran it live on a test box now (real gunicorn install, stock v2.219 + this branch):

- CA created with an `email` permitted constraint of `example.com`; a cert request with `email: user@outside.test` in the DN and no SAN is refused (`Names outside CA permitted NameConstraints: user@outside.test`), `jane@example.com` issues.
- `POST /api/v2/cas` with a `dns` value of `https://evil.com` returns 400 "not a valid dns name constraint", a non-ASCII value returns 400 with the IDNA hint, `.example.com` is accepted.

Box restored to stock afterward, `dpkg -V` clean, DB checksum unchanged.

Hemsby:
Pushed 6216b46.

1. **rfc822Name form semantics.** `_name_matches_subtree()` now distinguishes the three forms per RFC 5280 §4.2.1.10: `user@example.com` matches that mailbox only, `example.com` matches mail on that host only (not `sub.example.com`), `.example.com` matches the domain subtree only (not the bare host). The excluded side no longer over-blocks symmetrically. Domain is taken from the last `@` so a quoted local part is handled; cryptography already rejects bare multi-`@` addresses. Form-matching tests in `test_name_constraints_enforcement.py`.

2. **DNS leading dot.** Went with rejecting it at the `_normalize_name_constraints` boundary rather than changing the DNS match branch, per your lean. A bare `example.com` already covers the domain and its subdomains. The DNS branch itself is untouched (still reachable for constraints that come from an import rather than the wizard/API).

3. **Local part.** `_valid_email_name_constraint()` validates the mailbox local part as an RFC 5322 dot-atom (no leading/trailing/doubled dot, atext only, <= 64 chars); `a..b@x`, `.a@x`, `a.@x`, `a:@x` are rejected. Cases in `test_ca_name_constraints_api.py`.

CHANGELOG entry updated. Full backend suite on a Linux box: 4008 passed / 8 skipped. Verified live on the test box: `example.com` permitted refuses `user@sub.example.com`, `.example.com` permitted refuses `user@example.com`, dns `.example.com` and email `a..b@example.com` both 400 at CA creation.

NeySlim (approved):
Re-verified 6216b46 locally, all three points are properly addressed:

1. rfc822Name semantics now match RFC 5280 and OpenSSL. Checked the three forms against `user@example.com`, `other@example.com` and `user@sub.example.com`, then ran `openssl verify` on every certificate that was issued: `example.com` issues only the bare-host mailboxes, `.example.com` only the subdomain one, `user@example.com` only that mailbox, and each issued certificate verifies OK. The excluded side no longer over-blocks (`example.com` excluded still lets `user@sub.example.com` through), and the subject DN `emailAddress` check from the previous round still holds.
2. The leading dot is refused for `dns` at the API boundary, as agreed.
3. The local part is validated as a dot-atom: `a..b@`, `.a@`, `a.@`, `a:@` and a 65-character local part are 400, `first.last+tag@example.com` is accepted.

Full backend suite green here as well (4007 passed, 9 skipped), CI green across the matrix. Approving; I will resolve the `[Unreleased]` CHANGELOG overlap with #319 at merge time.

Issue #226, opened 2026-07-23
Issues with cert Templates


j2r2k2:
When you create a template, whatever template type you pick nothing changes in the KU or EKU

When you go in Certificate then issue, Cert type is here again? Totally makes sense if you don't pick a template here but if you pick a template, it is already linked that to a template. it should be there only if you don't select a template or been grey out and con't be modified if you select a template.

When you go in Certificate then issue, Cert type have not the same names as in templates menu? Custom and some others are not there also, which lead for example if you want to create a template for OCSP with only OCSPSigning EKU you can't issue a cert with only that EKU because a type is mandatory which includes EKU...

Missing a default OCSP Signing Template (as you propose the service in the App, it is legitimate that you have a default template for it)

Link to previous comment, a default EKU OCSPSigning check box is also something you should consider

NeySlim:
I confirm, this is a mess. Fixing.

NeySlim:
Thanks for the detailed report — all five points were confirmed and are now fixed on `dev` (will ship in the next release):

1. **Template type now drives the form** — changing the type in the template editor applies that type's Key Usage / Extended Key Usage / SAN presets.
2. **Templates now govern issuance** — when you issue from a template, its configured KU/EKU are what ends up in the certificate (previously the cert-type profile silently overrode them and always merged its own EKUs in). The *Cert type* selector is disabled when a template is selected, since the template defines it.
3. **New `custom` cert type at issuance** — imposes no EKU: only what you select is included, and with nothing selected the EKU extension is omitted entirely. Issuing a certificate with only `OCSPSigning` works now. Template types that have no direct issuance equivalent no longer silently fall back to the server profile.
4. **New “OCSP Signing” system template** (`digitalSignature` + `OCSPSigning`, 90-day validity per RFC 6960 practice). Existing installations receive it automatically on upgrade.
5. **`OCSPSigning` is selectable** in the template editor's EKU checkboxes.

Related fix that came out of this: the admin *Sign CSR* path also stripped `OCSPSigning`/`timeStamping` from CSRs, which blocked the usual delegated-responder workflow (key stays on the responder, operator signs its CSR) — operator-signed CSRs may carry these EKUs again; protocol enrollees (ACME/EST/SCEP) still cannot request them.

Verified end-to-end (template → issued certificate EKUs) and smoke-tested on SQLite and PostgreSQL.

j2r2k2:
[...]

PR #319, opened 2026-09-02
fix(certificates): keep the key size when the key type swaps its options (#318)

### LINKED ISSUE #318: Certificate template's EC curve is not applied to the issue form (Key Size left empty)
## What I'm trying to do

Issue an ECDSA certificate using a certificate template configured with Key Type `EC-P256`.

## What happens

When I select that template in the certificate issue form, **Key Type** is filled in as *ECDSA*, but **Key Size** is left empty. The curve from the template isn't carried across.

If I submit without noticing, issuance fails with an error saying the EC curve must be P-256, P-384 or P-521. Selecting P-256 by hand first works fine.

## What I expected

Selecting a template should populate both fields. If the template says `EC-P256`, the form should pre-select P-256 — the same way an RSA template fills in its key size.

## Environment

UCM 2.219 (`.deb`), Debian 13.

Hemsby:
Fixes #318.

## Problem

Selecting an EC certificate template in the Issue Certificate form fills Key Type as ECDSA but leaves Key Size blank, and submitting fails with `EC curve must be P-256, P-384, or P-521`. Picking the curve by hand first works.

## Cause

The Key Size `<Select>` rebuilds its option list (RSA sizes vs EC curves) whenever the key type changes. When that list swaps, the previously selected `<Select.Item>` unmounts and Radix fires `onValueChange('')`. The shared `SelectComponent` wrapper forwarded that as a real selection, so `key_size` was blanked immediately after `handleTemplateChange` (or the Key Type select's own onChange) had set it. The request then went out as `key_type=ecdsa` with `key_size=''`.

The Create CA wizard's key algorithm/size fields have the same pattern and the same latent bug.

## Fix

- `SelectComponent` now ignores an `onValueChange` whose value is not one of the current options. A genuine user pick is always on a rendered option, so this only drops Radix's spurious reset.
- `POST /api/v2/certificates` fills `key_size` from the template when the request sends a key type but no size, guarded so an explicit `key_type=rsa` never inherits the template's EC curve (defense in depth for API callers).

## Tests

- `frontend/src/pages/__tests__/issueCertificateTemplateKey.test.jsx` (new): EC template selection and a manual RSA -> ECDSA toggle both submit `key_size: '256'`. Both fail without the `SelectComponent` change.
- `backend/tests/test_issue_226_template_extensions.py`: EC template + blank size inherits the curve; RSA request + EC template falls back to RSA 2048.

Verified on a live instance: EC template now shows "NIST P-256 (ECDSA)" and issuance succeeds; the existing template-default behaviour (#226) is unchanged.

## Not in this PR

`TemplatesPage.jsx` `KEY_TYPE_OPTIONS` offers only `RSA-2048/4096` and `EC-P256/P384`, though `EC-P521` and `RSA-3072` are accepted everywhere else. Happy to add them here or separately if you want.


NeySlim (changes requested):
Thanks, the diagnosis is right and I could confirm the mechanism: it is Radix's hidden native `<select>` (`SelectBubbleInput`), keyed on the option set, which remounts when the Key Size list swaps and dispatches a change event with an empty value that the wrapper forwarded as a real pick. It only bites inside a `<form>`, which is why the issue form and the Create CA wizard are the two places. Re-ran the branch locally: the two new vitest tests fail with the wrapper from `dev` and pass with the change, the full frontend suite is green with the shared `SelectComponent` change (593 tests), and the backend template/certificate suites are green.

Requesting changes on two points:

1. **The approval path still has the old fallback (P2)**: the template fill in `cert_create.py` runs after the early return that creates the approval request, so a non-admin request is stored raw, and `_issue_approved_certificate()` in `policies.py` keeps the `if not key_type` logic with a `curve_map` that silently falls back to `SECP256R1`. Reproduced on the branch through `_issue_approved_certificate`: template `EC-P384`, `key_type=ecdsa`, `key_size=""` issues a `secp256r1` certificate with no error, while the same payload on the direct path issues `secp384r1`. Since the PR describes the backend fill as defense in depth for API callers, the approval path needs the same. Suggest factoring the "fill from the template, same algorithm only" logic into a helper (`utils/key_type.py` is a natural home) used by both routes, plus a case in `test_approval_template_issuance.py`.

2. **Non-string `key_type` now 500s (P3)**: the new `(key_type_in or 'rsa').strip().lower()` runs before `parse_issue_key_type()`, which used to `str()` the value. With a template, `{"key_type": 1, "key_size": ""}` (or a list) returns 500 "Failed to create certificate" on the branch versus 400 "Unsupported key type: 1" on `dev`. Wrap it in `str()` or reject non-string values with a 400 like the `cn` guards elsewhere. Same pattern as the `cn` note on #314, so let's not reintroduce it.

On `KEY_TYPE_OPTIONS` in `TemplatesPage.jsx`: agreed it should offer `RSA-3072` and `EC-P521`, but keep it out of this PR; a separate small PR is welcome.

Happy to merge once both are in.

Hemsby:
Pushed e2f6ca8 with both fixes.

1. **Approval path fallback.** Factored the "default key_type/key_size from the template, same algorithm only" logic into `fill_key_params_from_template()` in `utils/key_type.py` and wired it into both `cert_create.py` and `_issue_approved_certificate()`. The approval route now resolves the key through `parse_issue_key_type()` exactly like the direct route, so the `curve_map` with its silent `SECP256R1` fallback is gone. Your repro (`EC-P384` template, `key_type=ecdsa`, `key_size=""`) now issues `secp384r1` on both paths. New case in `test_approval_template_issuance.py` calling `_issue_approved_certificate` end to end.

   One consequence worth calling out: because the approval path goes through `parse_issue_key_type()` now, a stored request with a key size that was previously accepted only because `int()` parsed it (RSA 1024, 8192, ...) fails closed at approval time instead of issuing a weak or oversized key. `approve_request()` already surfaces that as `certificate_issued: false` with an `issue_error`, no crash, and the vote is preserved.

2. **Non-string key_type now 500s.** The `str()` coercion lives inside the shared helper, so `{"key_type": 1, "key_size": ""}` is back to a 400 "Unsupported key type: 1". Test in `test_issue_226_template_extensions.py`.

`KEY_TYPE_OPTIONS` (`RSA-3072` / `EC-P521`) held for a separate PR as you asked.

Verification: full backend suite on a Linux box against py3.13, 3974 passed / 8 skipped. eslint and the frontend suite unchanged from the last round (no frontend change in this push).

NeySlim (changes requested):
Re-verified e2f6ca8: the approval path now issues `secp384r1` for the template repro, the non-string `key_type` is back to a 400, and both suites are green here. The fail-closed behaviour for stored requests with an out-of-list RSA size at approval time is the right call (no weak or oversized key sneaks through, the vote is preserved), fine to keep.

One point remains, a consistency gap in the newly shared path:

1. **`curve` override is not recorded on the approval path (P3)**: `_issue_approved_certificate()` now resolves the key through `parse_issue_key_type(key_type, key_size, curve=data.get('curve'))`, but `template_overrides` is still computed from the raw `key_type` / `key_size` through the `_curve_to_pnum` / `effective_key_label` block, which never sees `curve`. Reproduced on the branch with an `EC-P384` template: approval with `key_type=ecdsa, curve=P-521` issues `secp521r1` and stores `template_overrides=[]`; the same request with `key_size=521` stores `['key_type']`, and the direct path with `curve=P-521` stores `['key_type']`. The certificate is right, it is only missing from the "diverges from template" filter. Fix as in `cert_create.py`: pass `key_type=normalized_key` to `compute_template_overrides()` and drop the now-dead label block. A case in `test_approval_template_issuance.py` with `curve` would lock it.

Merging as soon as that is in.

Hemsby:
Also ran it live on a test box now (deployed to a real gunicorn install, stock v2.219 + this branch): operator submits an EC-P384 template request with `key_type=ecdsa` and `key_size=""`, the request goes through the approval policy, admin approves, and the issued cert comes out `secp384r1` (source: approval). A non-string `key_type` on the direct path returns 400 "Unsupported key type: 1". Box restored to stock afterward, `dpkg -V` clean, DB checksum unchanged.

Hemsby:
Pushed b8bd60e. `_issue_approved_certificate()` now passes the generated `normalized_key` to `compute_template_overrides()` and the `_curve_to_pnum` / `effective_key_label` block is gone, so it matches `cert_create.py` exactly (this is the shape you flagged wasn't needed in the first round; with the shared parser now in place it is the clean way to do it). An approval with `key_type=ecdsa, curve=P-521` against an `EC-P384` template records `template_overrides=['key_type']`.

New case in `test_approval_template_issuance.py` with `curve`. Full backend suite on a Linux box: 3975 passed / 8 skipped. Also verified live on the test box through a real approval round-trip: the cert is `secp521r1` and `template_overrides` shows `key_type`.

NeySlim (approved):
Re-verified b8bd60e locally: an approval with `key_type=ecdsa, curve=P-521` against an `EC-P384` template now issues `secp521r1` and records `template_overrides=['key_type']`, the `key_size=521` and RSA cases are unchanged, and the template-default cases still record no override. Targeted suites green here (153 passed), CI green across the matrix, frontend unchanged since the first round where the full suite passed with the shared `SelectComponent` change.

Approving. The `[Unreleased]` CHANGELOG section overlaps with #317, I will resolve that at merge time.

PR #321, opened 2026-09-03
feat(templates): offer RSA-3072 and EC-P521 in the template Key Type dropdown

Hemsby:
Follow-up to #319 / issue #318, which NeySlim asked to keep as a separate small PR.

The template editor's Key Type dropdown offered RSA-2048, RSA-4096, EC-P256, EC-P384. This adds RSA-3072 and EC-P521.

Both were already fully supported below the UI: `_VALID_KEY_TYPES` in `api/v2/templates.py`, `parse_issue_key_type()`, and the RSA/EC generation in `cert_create.py` all accept them. Only the dropdown was missing them.

Verified end to end (new test in `test_issue_226_template_extensions.py`): a template with `key_type=RSA-3072` issues a 3072-bit RSA cert, one with `key_type=EC-P521` issues a `secp521r1` cert.

Deliberately out of scope, flagging so you can fold in or wave off:
- `PoliciesPage.jsx` keeps its own key-type list for the `allowed_key_types` rule (has EC-P521, not RSA-3072). That rule has no backend enforcement anywhere (grep of `allowed_key_types`), so it's inert today.
- `constants/config.js` has a dead `KEY_TYPE_OPTIONS` export nothing imports.
- ED25519 is in `_VALID_KEY_TYPES` but `parse_issue_key_type()` rejects it, so it would fail at issuance; not adding it here.

The dropdown renders raw values (`label: v`), so no i18n keys; `check:i18n` stays green.

CHANGELOG: `[Unreleased] > Added` (will conflict with #317/#319's entries the same way, whichever merges first).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ Starting with v2.48, UCM uses Major.Build versioning (e.g., 2.48, 2.49). Earlier
 
 ---
 
+## [Unreleased]
+
+### Added
+- The certificate template editor's Key Type dropdown now offers `RSA-3072` and `EC-P521`. Both were already accepted by the template API and produced working certificates at issuance; only the dropdown was missing them (#318 follow-up)
+
 ## [2.219] - 2026-09-01
 
 ### Fixed
--- a/backend/tests/test_issue_226_template_extensions.py
+++ b/backend/tests/test_issue_226_template_extensions.py
@@ -201,6 +201,28 @@ class TestTemplateDefaultsAtIssuance:
         assert isinstance(pub, rsa.RSAPublicKey)
         assert pub.key_size == 4096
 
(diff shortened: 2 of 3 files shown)


NeySlim (approved):
Verified on the branch: the two new values already flow through every consumer of `template.key_type` (issue form prefill, `parse_issue_key_type`, template override comparison, XCEP policy builder), and the new issuance test is green here as well. CI green across the matrix.

On the out-of-scope notes: agreed on all three. `allowed_key_types` being unenforced and the dead `KEY_TYPE_OPTIONS` export can wait; ED25519 sitting in `_VALID_KEY_TYPES` while `parse_issue_key_type()` rejects it is a trap worth closing in a follow-up (drop it from the accepted set unless we implement it).

Approving; the `[Unreleased]` overlap gets resolved at merge like the other two.

PR #339, opened 2026-09-07
feat(proxy): CIDR networks in UCM_TRUSTED_PROXIES + Helm chart wiring

Hemsby:
Closes the `UCM_TRUSTED_PROXIES` part of discussion #337 (reported by @szechyjs). The public/private port split you already fixed on `dev`; the HTTP-only admin listener you ruled out. This is the remaining piece plus the chart wiring.

## What changes

**CIDR networks in `UCM_TRUSTED_PROXIES`.** The value was matched against the immediate peer as an exact string (`*` the only wildcard), so behind an ingress controller or any proxy whose source address is not fixed, every allowed peer had to be listed one by one. Entries are now parsed as IP addresses or CIDR networks with `ipaddress`, and a network entry such as `10.42.0.0/16` trusts any peer inside it. Exact-IP and `*` behaviour is unchanged; the unset default is still loopback only.

**IPv4-mapped IPv6 peers.** A peer that arrives as `::ffff:10.42.0.5` (as a dual-stack `HOST=::` listener reports IPv4 clients) is matched against IPv4 entries and networks too. `PeerAddressNormalizer` already rewrites `REMOTE_ADDR` before this code runs, so this is belt-and-braces for any path that bypasses it, per your note.

**Unparseable entries.** An entry that is neither a valid address nor a valid network is dropped with a `warning` log line instead of being kept as a literal that can never match. It is never treated as trusted, so a typo cannot silently widen the set. If every entry is invalid, nothing is trusted (a loud failure), matching the old "what you set is what you get" behaviour.

The parse is `lru_cache`d on the raw env string, so it and the warning run once per distinct value rather than on every request.

**Helm chart.** New `proxy.behindProxy` (sets `UCM_BEHIND_PROXY=1`) and `proxy.trustedProxies` (sets `UCM_TRUSTED_PROXIES`, comma-joined), neither templated before. A chart deployment behind an Ingress can now honour `X-Forwarded-*` and proxy-injected client-cert headers without a hand-written `extraEnv` entry. Chart version `0.1.0` -> `0.2.0`. README gains a "Behind a reverse proxy" section; `values.yaml` documents that leaving `trustedProxies` empty trusts loopback only.

## Testing

- `backend/tests/test_trusted_proxy.py`: 9 new cases (CIDR in/out, mixed exact+CIDR, IPv6 CIDR, IPv4-mapped peer against IPv4 CIDR and exact, unparseable-entry-ignored-with-warning, all-invalid-trusts-nothing, compile helper). Full file plus `test_listen_address.py`, `test_mtls.py`, `test_public_endpoints.py` green on Linux / py3.13 (117 passed).
- `helm lint` clean; `helm template` emits no proxy env by default and both vars when `proxy.behindProxy=true --set proxy.trustedProxies={...}`.
- Live check on a real instance: with `UCM_TRUSTED_PROXIES=127.0.0.0/8,::1/128` and `UCM_BEHIND_PROXY=1`, a proxied login from loopback had its `X-Forwarded-For` honoured and recorded in the audit log as the forwarded IP, confirming the network match on the running server. Reverted after.

## Docs

`CHANGELOG.md` `[Unreleased] > Added`; `charts/ucm/README.md`; `docs/testing/PUBLIC-ENDPOINTS.md`.


NeySlim:
Merged, thanks. Reviewed against the three points from the discussion and probed a few extras (`*` inside a list is ignored rather than trusting everyone, host bits in a CIDR are masked, an unparseable peer is refused, a zoned IPv6 peer matches); all as expected, and the full suite is green on the merged tree.

One follow-up commit on `dev` on top of yours: an entry written in the mapped form (`::ffff:10.0.0.5`) now also matches the bare IPv4 peer the normalizer reports (with a test), your CHANGELOG entry carries the contributor credit, and the chart `appVersion` moves from 2.170 to 2.223 since the default image tag falls back to it. The wiki pages for mTLS and EST now mention CIDR entries as well.

PR #340, opened 2026-09-07
feat(templates): smartcard logon EKU/SAN in the editor + built-in template

Hemsby:
Closes discussion #336 (requested by @szechyjs). Per your go-ahead on the thread.

Most of a smartcard logon certificate already works from the Issue Certificate form: `msSmartcardLogin` is in the EKU catalog and the `UPN` SAN type encodes the Microsoft UPN OID. The gap was the template editor, whose EKU and SAN checkbox lists are fixed short sets, so a reusable Smartcard Logon template could not be assembled even though the template API and the issuance path both accept the values.

## What changes

**Editor.** `msSmartcardLogin` is added to the Extended Key Usage list and `upn` to the SAN types. A `smartcard_logon` template type is added, and selecting it applies `clientAuth` + `msSmartcardLogin` with a UPN SAN, the same way the other types apply their presets.

**Built-in template.** A Smartcard Logon system template ships next to the others: `clientAuth` + `msSmartcardLogin`, UPN SAN, RSA-2048, 397 days, subject CN `{username}`. It is seeded in both places system templates are defined, per your note:
- `SYSTEM_TEMPLATES` in `backend/services/template_service.py`
- the startup seed in `backend/app.py` (the fresh-install list, plus the post-install catch-up, which was a single `if not "OCSP Signing"` check and is now a name list, so an upgraded instance also picks it up)

`san_types` on a template is only a preview hint today, as you said, so the `upn` entry there is about discoverability, not enforcement.

**API.** `smartcard_logon` is accepted as a `template_type`.

**i18n.** The SAN type checkboxes now render the Issue Certificate form's existing localized labels (`certificates.sanDns` ... `certificates.sanUpn`, already present in all nine locales) instead of a bare upper-cased token, and the section heading uses `common.subjectAltNames`. No new translation keys. The EKU checkboxes still render the raw token (`serverAuth`, `clientAuth`, ...), so `msSmartcardLogin` shows as-is, matching the existing rows and the Key Usage list. If you would rather it read "Smartcard Logon", that is one key across nine locales and I will add it in this PR or a follow-up, your call.

## Testing

- `backend/tests/test_templates.py`: two new cases (the built-in template ships with the right EKU/SAN; `smartcard_logon` is accepted over the API). Full `test_templates.py` plus `test_issue_226_template_extensions.py`, `test_wstep_template_eku.py`, `test_template_overrides.py` green on Linux / py3.13 (115 passed).
- Built the frontend and exercised it on a real instance: the built-in template appears with `clientAuth` + `msSmartcardLogin` in its detail panel; the editor's type dropdown lists Smartcard Logon; picking it ticks digitalSignature/keyEncipherment, clientAuth, msSmartcardLogin and UPN; the SAN labels render as DNS Name / IP Address / Email / URI / UPN. Verified the post-install catch-up creates the template on an existing database at startup.

## Docs

`CHANGELOG.md` `[Unreleased] > Added`.


NeySlim:
Merged, thanks. Checked end to end on the merged tree: a certificate issued from the built-in template through the API carries `clientAuth` + `1.3.6.1.4.1.311.20.2.2`, the UPN as the Microsoft otherName (`1.3.6.1.4.1.311.20.2.3`, UTF8String), and digitalSignature + keyEncipherment; the post-install catch-up also seeded the template on an existing database at startup. On the EKU label: keep the raw `msSmartcardLogin` token, consistent with the other rows of that list. The wiki template page now lists the built-in template and the two editor options.

PR #351, opened 2026-09-13
feat(export): make root CA inclusion optional

B0F1B0:
## Summary

- add an explicit `include_root` export option, defaulting to `false`
- export leaf + intermediate CA certificates by default when the CA chain is enabled
- allow callers to opt in to packaging the self-signed root CA
- apply the same chain semantics to PEM, PKCS#7, PKCS#12/PFX, and JKS exports
- expose the option in the certificate export dialog and translate it in all supported locales

## Motivation

TLS servers normally send the leaf certificate and intermediate CA certificates, while the self-signed trust anchor is installed in the client's trust store. Previously, enabling the CA chain did not provide a way to exclude the root consistently across export formats.

The resulting behavior is:

| `include_chain` | `include_root` | Exported certificates |
| --- | --- | --- |
| `false` | ignored | Leaf only |
| `true` | `false` | Leaf + intermediate CA certificate(s) |
| `true` | `true` | Leaf + intermediate CA certificate(s) + root CA |

Inline certificate chains are split before export so these options are honored consistently. API callers that need the previous full-chain packaging can set `include_root=true`.

## Screenshot

UCM export dialog with optional Root CA checkbox highlighted

## Testing

- 7 targeted backend export regression tests passed
- 152 frontend service tests passed
- all 9 locale files passed the i18n synchronization check
- production frontend build completed successfully (9,565 modules)
- validated on a UCM 2.229 Debian installation


NeySlim (changes requested):
Thanks, the product direction is right and the normal chain cases work: I verified PEM, PKCS#7, PKCS#12/PFX and JKS with a root -> intermediate -> leaf hierarchy, and each export contains 2 certificates by default and 3 with `include_root=true`. The certificate suite, frontend suite, i18n sync and frontend build are green.

Requesting changes on six points before merge:

1. **The shared default breaks deployment fullchains (P1)**: `_build_ca_chain()` now defaults to `include_root=False`, but `DeployService._chain_pem()` calls it without passing that argument. This changes a non-export caller as a side effect. Reproduced with a root-issued leaf: `resolve_files()` produces a `fullchain_path` containing only the leaf, and the existing `test_fullchain_includes_issuer` fails on this branch while passing on `main`. The full backend suite stops on that regression after 2010 passing tests. Either preserve the old helper default and pass `include_root=False` explicitly from export endpoints, or make the deployment behavior an explicit, tested decision rather than changing it implicitly.

2. **Issuer reconstruction can export an unrelated CA with the same DN (P1)**: `_resolve_issuer_cert()` compares `cert.extensions...key_identifier.hex()` with `CA.ski`, whose stored representation does not match that value, then falls back to `.first()` by subject DN without verifying the signature. I reproduced this with two managed intermediate CAs sharing the same subject: an exported leaf signed by one CA carried the other CA in its chain. The resolver must normalize SKI/AKI consistently and cryptographically verify each candidate before accepting it. The existing helpers in `utils.cert_issuer` already provide the right primitives.

3. **Self-issued is treated as self-signed (P1)**: both the chain walk and final root removal use only `subject == issuer`. A rollover/cross certificate can be self-issued while being signed by another key, so `include_root=false` removes a certificate that is not a trust anchor. Reproduced with such a certificate inline in the leaf chain. Use a real signature check such as `is_self_signed()` from `utils.cert_issuer` before stopping the walk or dropping the last certificate.

4. **The new checkbox is ineffective on the Account page (P2)**: `ExportModal` returns `includeRoot`, but `userCertificatesService.export()` does not destructure or send it. The PKCS#12 branch through `accountService.downloadMTLSCertificate()` also omits it. The UI therefore offers "Include Root CA" while both account export paths always use the backend default. Please thread the option through both services and add a page or service test that asserts the request body.

5. **JSON boolean coercion is unsafe (P2)**: `include_root = bool(data.get('include_root', False))` makes `{"include_root": "false"}` evaluate to `True` and unexpectedly includes the trust anchor. Require an actual JSON boolean and return 400 for other types. The same validation should cover the related boolean export options while this payload is being touched.

6. **The API behavior change is undocumented (P2)**: `docs/API_REFERENCE.md` still labels the chain example as "Full chain" and does not describe `include_root` or its new default. This is a compatibility-visible default change for existing API consumers, so the API reference and CHANGELOG need to state it explicitly.

Please also retarget the PR to `dev`; releases flow from `dev` through the release branches rather than accepting feature work directly into `main`.

Happy to re-review once these are addressed.

B0F1B0:
Thanks for the detailed review. I addressed all six points in `100d181` and rebased the branch onto the current `dev` head (`f661ba1`):

1. Preserved the historical `_build_ca_chain(..., include_root=True)` default so deployment fullchains remain unchanged; export callers now opt out explicitly.
2. Normalized AKI/SKI values and cryptographically verified issuer candidates, including an adversarial same-DN CA test.
3. Replaced subject/issuer equality checks with `is_self_signed()`, including a self-issued rollover regression test.
4. Threaded `includeRoot` through both Account page export paths and added page/service/backend coverage.
5. Added strict boolean validation for JSON and query export options; ambiguous string values are rejected.
6. Updated the API reference and changelog.

The PR is now targeted at `dev`. After the latest rebase, 14 targeted backend tests and 163 targeted frontend tests pass. The production frontend build, i18n synchronization and lint (0 errors) also pass.

Could you please re-review when convenient?

NeySlim (approved):
Re-verified `100d181` locally, all six review points are properly addressed:

1. `_build_ca_chain()` keeps its historical `include_root=True` default for non-export callers, while every export path passes the requested value explicitly. The deployment fullchain regression test is green again.
2. AKI/SKI values are normalized before shortlisting and every managed or trust-store candidate must verify the certificate signature. The homonymous-CA adversarial case now resolves the real issuer.
3. Chain termination and root removal use `is_self_signed()`. The self-issued rollover case keeps the cross-certificate and removes only the actual trust anchor.
4. `includeRoot` now reaches both Account page paths, including the dedicated mTLS PKCS#12 endpoint, with component and service coverage.
5. Certificate, user-certificate and mTLS export endpoints now require real JSON booleans and strict `true`/`false` query values. The previous string `"false"` repro returns 400.
6. The API reference and CHANGELOG describe the default and opt-in behavior, and the PR now targets `dev`.

The targeted backend set is green (189 passed), as are the focused frontend tests (163 passed). Full backend suite: 5033 passed, 14 skipped. Full frontend suite: 617 passed. Ruff, ESLint, i18n synchronization and the production build are green as well.

Approving.

PR #360, opened 2026-09-19
docs(scep): add a Microsoft Intune enrollment guide

stefanelul2000:
## Summary

- add an end-to-end guide for issuing device certificates through Intune's live SCEP challenge validation
- cover the Entra app registration, its two application permissions and admin consent, the UCM profile fields that enable validation, publishing the endpoint, and the Intune-side certificate profile
- document the trust chain per platform, since Intune's Destination Store field is Windows only
- link the guide from `docs/README.md` under a new Integrations section

## Problem

UCM has supported Intune SCEP challenge validation since v2.209, but there was no documentation for it. Intune does not accept a static challenge password — it issues a per-device, per-request encrypted blob that only its own API can validate — so the setup is not something an operator can infer from the SCEP page. It requires an Entra app registration with two specific application permissions, admin consent, and auto-approve forced on, and gives an unhelpful failure at the service-discovery step if any of that is missed.

Two further details cost real troubleshooting time and are now written down:

- A SCEP failure is not an HTTP failure. Per RFC 8894 §3.3.2 protocol errors travel inside a signed PKI message under HTTP 200, so `access.log` shows a rejected enrollment as an unremarkable `200`.
- Requests refused during validation are rejected before a request record is created, so they never appear on the SCEP Requests page. An empty page does not mean the device never arrived.

## Documentation

`docs/integrations/intune-scep.md` — 807 lines, 21 screenshots, structured as Prerequisites, five steps and Troubleshooting.

- **NGINX exposure is TLS passthrough** (`stream` + `ssl_preread`), so NGINX holds no key. What that costs is stated rather than left to be discovered: UCM's own certificate is what devices see, path filtering goes with the HTTP layer so the whole instance answers on that hostname, and client IPs in the logs become the proxy's address. It also corrects the assumption that this matches App Proxy — App Proxy terminates at Microsoft's edge and the connector opens a second connection inward, which is bridging, not passthrough.
- **Deploying the root CA to the Intermediate store as well on Windows** is written up as a field workaround, not a requirement, since Microsoft documents one trusted certificate profile per CA certificate and does not ask for it. The guide says what it costs (nothing) and how to skip it.

## Verification

No code changed, so no tests. What was checked:

- three claims in the Intune profile section verified against the source rather than asserted — the SCEP path copies the CSR's `SubjectAltName` through verbatim so the `IntuneDeviceId` URI SAN arrives unchanged; UCM only builds DNS, IP and email name-constraint subtrees so a URI SAN is never constrained; and the bundled Client Certificate template already asserts `digitalSignature` + `keyEncipherment` with `clientAuth` over 365 days, which is what the profile asks for
- log locations checked against `packaging/debian/ucm.service` (`StandardOutput=journal`, `SyslogIdentifier=ucm`) and `backend/app.py`, and both destinations documented because which one is live depends on whether the rotating file handler opened its file at startup
- all 21 image references and all relative links resolve
- screenshots reviewed for credentials: tenant and client IDs redacted, the client-secret screenshot captured before the secret existed, the App Proxy tenant hostname redacted


NeySlim:
Thanks, this is the guide the feature was missing, and the checks you list in the description (SAN passthrough, name constraints, the bundled template) match the code. Five things before I merge it, the first one being the one that matters:

1. **SCEP Server URL for Windows.** The "Common to both platforms" table gives the URL ending in `/pkiclient.exe` for both platforms. As you found in #228, the Windows client appends `/pkiclient.exe` itself, so the Windows profile has to take the URL without it, and iOS/iPadOS with it. Worth its own row, or a line in the Windows section.
2. **Windows enrollment is not working yet.** Your log line in #228 pinned it: the Windows client wraps the session key with RSAES-OAEP and UCM only accepts PKCS#1 v1.5. The fix is on its way; I would rather merge the guide with that release than before it, so either wait for it or add a one-line note in the Windows section pointing at #228 for now.
3. **Root CA in the Intermediate store.** You tested that one in #228 and it changed nothing, and the text already says Microsoft does not ask for it. I would drop the row and the paragraph rather than keep a workaround that has not earned its place.
4. **Three `TODO screenshot` comments** are still in the file. Fill them or remove them; an HTML comment in a guide tends to stay there for years.
5. **The "Secret storage" note** names two internal modules and states how the static challenge is stored. That paragraph came from a code comment, and it belongs there rather than in an operator guide; the sentence "the client secret is encrypted at rest" is all the reader needs.

I will add the CHANGELOG entry on merge. No need to squash anything on your side.

stefanelul2000:
Thanks — all five done, squashed into `b4f05ca2` on top of the original commit.

**1. SCEP Server URL for Windows.** The shared row no longer carries a URL; it points at a per-platform table in the note below it, iOS with `/pkiclient.exe` and Windows without. Added the symptom too, since the doubled `.../pkiclient.exe/pkiclient.exe` is what a reader actually sees if they get it wrong.

**2. Windows enrollment.** One-line note at the top of the Windows section pointing at #228, plus that iOS/iPadOS is unaffected and the settings below are still correct to configure.

**3. Root CA in the Intermediate store.** Dropped: row, rationale, checklist line and the placeholder. Windows is back to two profiles.

One consequence worth flagging: `intune-scep-profile-trustedCA.png` showed all three Windows profiles, so it contradicted the text once the third was gone and I removed it as well. The iOS list screenshot is unaffected. Happy to re-take the Windows one with two profiles and put it back if you want the illustration.

**4. TODO screenshots.** All three removed.

**5. Secret storage note.** Cut to "the client secret is encrypted at rest" — the module names and the comparison with the challenge-password helper are gone.

Also rebased onto current `dev`.

NeySlim:
Merged, thanks. Two small things done on my side in the same push: the CHANGELOG entry crediting you, and the Windows note now reads "Windows needs UCM v2.232 or later" instead of "does not complete yet", since the OAEP fix from #228 ships in the same release as this guide. The wiki's SCEP page points at the guide too.

Yes to the Windows trusted-certificate screenshot with two profiles whenever you have a minute; a one-image PR is fine.

PR #361, opened 2026-09-19
Read UCM's own logs from the web UI

stefanelul2000:
## Read UCM's own logs from the web UI

Adds an **Administration → System Logs** page that reads back what UCM wrote, without shell access to the host. Today the only route to those lines is a file under `/var/log/ucm/`, or `docker logs` in a container — neither of which an operator running UCM as an appliance necessarily has. `journalctl` is not a third route: the application log never goes to stdout on a native install, so the journal has never carried it.

It is the log that explains a failure no record was ever created for. A SCEP enrolment refused during validation is rejected before a request row exists, so the Requests page stays empty and the reason lives only here.

### Screenshots

chrome_Q4TebgA9cI
chrome_zqtSIWG8ud
chrome_zEOIUooS5a
chrome_5Mz0PCOvhJ
chrome_28u9UERb2k

### What's in it

**A canonical log file (`backend/utils/app_log.py`, new).** UCM had no single path it could read back on every deployment. Native installs keep `/var/log/ucm/ucm.log` first so logrotate and operator tooling are untouched; Docker, which has no `/var/log/ucm`, gets `DATA_DIR/ucm.log`. Rotation is bounded and configurable through `UCM_LOG_MAX_BYTES` / `UCM_LOG_BACKUPS`, and `UCM_LOG_FILE` overrides the native path. Under `TESTING` it still goes to stderr. The dead `AUDIT_LOG_FILE` setting is removed.

**A read endpoint (`GET /api/v2/system/logs`).** `admin:system`, same gate as the bundle download.

```
?source=app|access|error|journal   default app; only sources that exist are offered
&lines=200                         default 200, max 5000
&level=DEBUG|…|CRITICAL            a floor, not an exact match; default INFO
&component=services.scep           the logger subtree
&q= &exclude=                      substring over message + logger, or patterns with:
&regex=true
&since= &until=                    server local time
```

The response carries `components`, `available_sources`, `levels` (counts over everything matched), `timezone`, `truncated` and `scan_truncated`.

Three line formats are parsed: UCM's own, and gunicorn's error and access lines. A line matching none of them continues the record above it, so a traceback stays one entry rather than forty; secrets are redacted before anything leaves the process.

**The page (`SystemLogsPage.jsx`).** `ResponsiveLayout` + `ResponsiveDataTable`, matching Audit Logs: one row per record with time, level, component and message; newest first; multi-select with copy-all / copy-selected in the log's own shape; live tail; a filter panel; Quick Help and Guide in all nine locales.

**Packaging.** `ucm` joins `systemd-journal` in the DEB postinst and the RPM spec, and the RPM unit states `StandardOutput/Error=journal` with `SyslogIdentifier=ucm`.

**This changes what can be read, not what gets written.** Nothing here reroutes a single line into the journal. On a native install every stream UCM produces already goes to a file:

| Stream | Native destination | In the journal? |
|---|---|---|
| UCM application log | `/var/log/ucm/ucm.log` | no |
| gunicorn access | `/var/log/ucm/access.log` | no |
| gunicorn error | `/var/log/ucm/error.log` | no |
| systemd lifecycle, direct stdout/stderr writes | — | yes |

So `journalctl -u ucm` answers "did it start, did it crash", and nothing about why an enrolment failed — ask it about a SCEP failure and you get silence, while the line sits in a file an appliance operator cannot reach. That is the problem this PR exists to solve, and it is solved by making the file readable from the UI rather than by writing everything twice.

The group grant fixes reading. `collect_journal` shells out to `journalctl -u ucm` as the `ucm` user, who could not read the system journal, so the *existing* diagnostic bundle has been shipping an empty journal section on essentially every install. The UI's Journal source runs the same command and benefits the same way. `StandardOutput/Error=journal` are systemd's defaults, stated so the bundle is not collecting from an implicit destination; `SyslogIdentifier=ucm` tags those entries as the DEB unit already did.

### Decisions worth a second opinion

**Reads are deliberately not audited.** `AuditService` echoes every entry to the application logger, so auditing a read of that log writes a line into the log being read, which the next read shows, for ever. Live mode polls, so the viewer would bury the lines it exists to surface under a record of itself. The bundle download stays audited: it is rare, deliberate, and leaves with a copy of the file. The reasoning is in a comment at the endpoint.

**Live mode polls every five seconds rather than streaming.** The WebSocket infrastructure exists, but a log line pushed over the event bus is itself logged by the push — a loop an interval cannot make.

**The component list is every subsystem UCM has, not just what is on screen.** Built from `Logger.manager.loggerDict` filtered to UCM's own top-level packages, unioned with the loggers present in the lines read. A component that has been quiet is still selectable; picking one takes everything beneath it, so `services` matches `services.scep.scep_service`.

**An empty journal is not offered.** `journalctl` exits 0 and prints `-- No entries --` when the unit has never run or the service user cannot read the system journal. Offering it reads as an empty log rather than as one that does not exist here.

**A bad regex matches nothing rather than 500s.** Half-written patterns are the normal state of a filter box someone is typing into.

### Tests

| | File | Tests |
|---|---|---|
| Backend | `test_log_reader.py` | 72 |
| Backend | `test_system_logs_endpoint.py` | 18 |
| Backend | `test_app_log_path.py` | 12 |
| Backend | `test_journal_group_is_granted.py` | 7 |
| Frontend | `SystemLogsPage.test.jsx` | 25 |
| Frontend | `systemLogsI18n.test.js` | 19 |
| | **Total** | **153** |

Both full suites were run locally and pass. Every commit also passes its own tests when checked out alone, so the branch stays bisectable.

### One thing I could not verify

**The journal source has only ever run against a faked `journalctl`.** My dev host is WSL with no `ucm` unit, so the journal never appears in the picker there — correctly, since an empty journal is not offered, but that means the reading path itself is covered by unit tests and not by a live run. Someone on a real DEB or RPM install should confirm it shows up and reads.


NeySlim:
WHAT A GREAT IDEA !

NeySlim:
Thanks, this is a feature I want, and the shape is right: one file UCM can read back, a bounded tail, one gate, redaction shared with the bundle. 153 tests run and pass here, the nine locales are in sync, and it merges cleanly on `dev`. It goes into 2.233: 2.232 is at rc2 and this touches the logging chain of every deployment plus both packagings, so it will get its own release candidate.

Six things to change before merge, one of which you flagged yourself.

**1. The Journal source collapses into one record.** `journalctl --output=short-iso` writes `2026-09-19T21:30:08+02:00 host ucm[409576]: message`, and none of the three patterns in `parse()` matches that, so every journal line is taken as the continuation of the one above. Against 40 real lines from a DEB host, `parse()` returns one record of 40 lines, no timestamp, no level. `test_reads_the_journal` cannot see it: it feeds a single line and checks a substring. A fourth pattern for short-iso (timestamp, identifier, message; level None) and choosing the pattern per source (app, error, access, journal) instead of trying all three everywhere fixes it, with a test that N journal lines give N records each carrying its `ts`.

**2. `journalctl` runs on every request, and writes into the log being read.** `read()` builds `available_sources()` each call, which calls `journal_text()`, which shells out; twice when the source is the journal; every 5 s in live mode and on every keystroke of the search box. On a host where `ucm` is not yet in `systemd-journal` (source installs, or until the service restarts after upgrade) `journalctl` exits 1 with "No journal files were opened due to insufficient permissions", not 0 with `-- No entries --`, and `collect_journal` logs that at INFO into `ucm.log`, so the viewer feeds itself a line every 5 s, the loop you avoided for the audit. Cache the journal's availability (once per process or a short TTL), do not probe it when another source is asked for, and log the probe at DEBUG.

**3. `regex=true` can freeze the server.** gunicorn runs one gevent worker; a user pattern is searched over up to ~15,000 records with no bound. Measured: `(a+)+b` on a 29-character line takes 15 s, and `^(\w+\s?)*$` about 14 s per line. Admin-only, so a self-inflicted outage, but ACME, SCEP and OCSP stop answering with it. Either drop regex from this PR (substring plus exclude covers the use) or keep it with a hard pattern length and say so in the help.

**4. Redaction misses UCM's own secrets.** The shared pattern is `\b(pass(word|wd)?|secret)`, and `\b` does not match after an underscore, so `client_secret=…` (Intune), `challenge_password=…` (SCEP), `bind_password=…`, `smtp_password: …`, `"password": "…"` in JSON, `Cookie:`, and `https://user:pw@host` all pass through. Pre-existing in the bundle, but this PR turns it into a live view whose help promises server-side redaction. `(?<![A-Za-z0-9])` instead of `\b`, an optional quote around the separator, plus `cookie` and URL userinfo, each with a test that dies when the pattern is loosened.

**5. Journal access is a privilege grant to weigh.** `usermod -aG systemd-journal` gives the service account the whole host journal (sshd, sudo, every unit), for a source that, as you write, only says whether the service started. The endpoint is bounded (`-u ucm` fixed, no shell, timeout) but any code running inside UCM now reads the host's journal. If the source stays, I would rather see `SupplementaryGroups=systemd-journal` in the two units, declarative and revocable, than a `usermod` in postinst and `%post`.

**6. No CHANGELOG entry**, and the added texts carry em dashes (145 lines across the eight help files, `helpGuides.js`, one `uk.json` string, and the `'—'` level badge in the page). The project keeps those at zero in public text; a colon or a shorter sentence does the job, and the badge can be empty.

Smaller ones, a few lines each: the copy button writes `ts LEVEL [logger] message` while the file writes `ts [logger] LEVEL message`, so a pasted line does not parse back; the docstring of `test_system_logs_endpoint.py` says every read is audited, the endpoint and its own test say the opposite; the API without `level` returns everything while the description says INFO is the default; `UCM_LOG_MAX_BYTES`, `UCM_LOG_BACKUPS` and the Docker `DATA_DIR/ucm.log` are not in the admin or Docker docs, and the docstring in `app_log.py` cites `/var/lib/ucm` for the RPM, which the spec does not define.

Verified without finding: the `admin:system` gate, the whitelisted sources, `lines` bounded 1..5000, the 2 MB tail, tolerant `since`/`until`, a bad regex matching nothing, `/var/log/ucm/ucm.log` unchanged on native installs, `docker logs` kept, `AUDIT_LOG_FILE` with no reader left, the postinst and `%post` idempotent, and the bundle's journal section finally filled.

stefanelul2000:
All 6 are in, plus the smaller ones and one you did not mention. 11 commits on top of what you reviewed: one per point, with the help text and the changelog split out of point 6, 2 for your smaller list, and 1 for something you did not ask for.

### 1. Journal collapsed into one record

Exactly as you describe: none of the 3 patterns matched short-iso, so the whole file continued the first line.

There is a fourth pattern now, and `parse()` takes the source instead of trying every pattern everywhere, so app, access, error and journal each get the format they are actually written in. The error stream is the only one with 2, since gunicorn writes its own lines there and an unhandled traceback reaching its logger arrives in UCM's format.

N journal lines now give N records, each with its `ts`, the identifier stands in for the component, and no level is invented. You were right about `test_reads_the_journal` too: it fed 1 line, so a parser that folds every line into the one above looked identical to one that works. It feeds 2 now and reads them back. Against 400 real journal lines from a host here: 400 records with 400 timestamps, where the old parser gave 1 record and no timestamp.

### 2. journalctl on every request

Availability is cached for 300 seconds, the probe is logged at DEBUG, and reading the journal doubles as its own probe, so 1 request never shells out twice.

One thing I did not do literally: *do not probe it when another source is asked for*. If the journal is never probed while the application log is being read, it can never appear in the source list, so nobody could select it. The TTL is the other half of your sentence, and it turns per-keystroke probing into at most 1 probe every 5 minutes. It is one constant if you want it tighter, or once per process.

### 3. regex

Dropped. A length cap does not fix it: `(a+)+b` is 6 characters, and your 15 seconds is 15 seconds of no ACME, SCEP or OCSP.

The parameter is refused with a `400` rather than ignored, since no release ever answered it, and the help no longer offers it. If it returns it should be on an engine with a linear time guarantee, not behind a length limit.

### 4. Redaction

Your diagnosis was half of it. I took the names out of the settings and the provider credential schemas instead of guessing, and tested each in 3 forms. The prefixed cases you named leaked, and so did `secret_key`, `eab_hmac_key`, `user_pin` and `passphrase` whatever their shape, plus any quoted value holding a space: `"password": "hunter 2"` left the ` 2` behind.

That was 21 of 87 forms before and is 0 of 108 now, across 36 names. 14 settings that only read like secrets, among them `password_set`, `min_password_length`, `token_label`, `token_url`, `key_type` and `has_private_key`, are pinned as staying legible, because redaction that eats the diagnostics is no better than redaction that misses. Cookie headers go whole rather than to the first semicolon, since a jar holds several.

One cost: redaction over the 2 MB tail went from 0.11 s to 0.25 s per request, measured here.

### 5. Journal access

`SupplementaryGroups=systemd-journal` in both units, `usermod` out of postinst and `%post`. Nothing released ever ran that usermod, so there is no membership to clean up.

One behaviour difference worth stating: if the group is missing the unit refuses to start, where the usermod quietly did nothing. systemd ships that group in its own sysusers, so it exists wherever these units run.

On whether the source earns the grant: the grant is there for the bundle, which collects the journal either way, so dropping the page's journal source would remove a dropdown entry and keep the privilege. And the journal holds more than *did it start*. `migration_runner` prints migration progress to stdout, `app.py` prints the initial admin notice, and systemd writes the restart, exit code and OOM records. None of that reaches `ucm.log`.

### 6. CHANGELOG and em dashes

Entry under `## [Unreleased]`, split across Added, Changed and Fixed. 

Em dashes are at 0 across the 8 help files, `helpGuides.js`, `helpContent.js`, the uk string and the badge, which is now empty for a line that carries no level rather than a placeholder that reads as a level of its own. 2 tests hold that: no `logs` string in any locale carries one, and no help text does. The help modules are imported rather than read as text, so a broken template literal fails the suite as well.

### Smaller ones

- Copy writes `ts [logger] LEVEL message`, the file's own order, pinned by a test.
- The docstring of `test_system_logs_endpoint.py` now says what the endpoint does.
- The API defaults the level to `INFO`, so a caller that omits it reads the log the help describes.
- `UCM_LOG_FILE`, `UCM_LOG_MAX_BYTES` and `UCM_LOG_BACKUPS` are in the admin guide, the rotation page and the Docker page, with the container's `/opt/ucm/data/ucm.log` beside the stdout stream `docker logs` still shows.
- `app_log.py` names `/opt/ucm/data` for the RPM, and I corrected the same wrong claim sitting next to it in `settings.py`.

### One you did not ask for

Nothing debounced the search box, so every character cost a tail read, a redaction pass and a parse, which the redaction work above made dearer. Both free-text boxes now wait 300 ms for the typing to stop, the same pause `SearchBar` uses elsewhere. Typing `scep` costs 1 request instead of 4.


### Tests

263 across the 4 log files, 48 in the frontend, `check-i18n-sync` clean on all 9 locales. The full backend suite I leave to CI.

NeySlim:
Merged, with thanks: every point checked in behaviour on my side (the journal read back record by record, the probe cached, `regex` answered 400, the nine secret shapes I listed all masked, the units carrying the group), 263 backend and 50 frontend tests here, and the page running on a test install.

Two notes. The CHANGELOG conflicted with the rc3 entries, so I merged locally; your four entries are in, shortened to the two sentences the file keeps per entry and credited to you. And it ships in 2.232 after all, through a further release candidate, because the licence change deserves it in the same release as the Intune work; the packaging change gets validated on real DEB and RPM upgrades before the stable.

PR #362, opened 2026-09-20
feat(deploy): add CRL deployment via SSH/SFTP and per-binding reload commands

B0F1B0:
### Push CRLs to deployment targets

- add CRL deployment bindings to **CRL & OCSP > CRL Deployment**
- push a newly generated or changed CRL over SFTP instead of requiring every target to periodically pull it with cron
- support a single DER CRL or a PEM bundle containing the selected CA CRL followed by its parent CRLs
- keep delivery durable and isolated per target: a failure on one target does not prevent delivery to another, and failed attempts remain queued with backoff
- allow CRL bindings to be deployed on demand and edited after creation

### Keep connection settings separate from deployment settings

**Settings > Deployment** now defines only the reusable SSH/SFTP connection: name, host, port, SSH user and private key.

Destination paths, file options and the optional **reload command** now belong to the concrete binding:

- certificate paths and reload commands are configured in a certificate's **Deployment** section
- CRL path, format, parent-CRL option and reload command are configured in **CRL & OCSP > CRL Deployment**

This separation matters because one SSH target may host several services. Certificates and CRLs can require different destination paths and **different** reload commands even when they are delivered to the same machine. Keeping these values on each binding makes the target reusable and prevents an unrelated global command from running for every deployment.

Existing target-level reload commands are copied to existing certificate and CRL bindings by migration 091. The legacy database column is retained for downgrade compatibility but is no longer exposed or executed.

### Binding updates and UX

- add editing for certificate and CRL bindings
- immediately perform the first delivery attempt after a binding is saved; only failed attempts enter the normal retry schedule
- allow optional certificate paths to be cleared while editing
- hide **Include parent CRLs** for DER, because DER represents one CRL rather than a PEM bundle
- leave a new CRL destination path empty and show `/root/certs/CRL.crl` only as an example placeholder
- preserve the existing per-binding root-CA option for certificate fullchains

## Why

Before this change, UCM could push renewed certificates, but CRL consumers had to discover changes themselves. A typical reverse proxy therefore needed a cron job which downloaded multiple CRLs, converted DER to PEM, assembled a bundle, validated the proxy configuration and reloaded the service.

CRL deployment lets UCM notify the systems that actually consume its revocation data. The same audited SSH/SFTP delivery mechanism used for certificates can push the complete updated CRL and run the binding-specific validation/reload command. This reduces polling delay and keeps deployment state, failures and retries visible in UCM.

### Before

The reload command was part of the global SSH target:

Before: Settings Deployment target contains a global reload command

Certificate bindings contained only destination paths:

Before: certificate Attach target dialog without a binding-specific reload command

### After

The target contains only its reusable SSH/SFTP connection:

After: Settings Deployment target contains only SSH connection settings

Certificate-specific paths and the reload command are configured on the certificate binding:

After: certificate Attach target dialog with binding-specific paths and reload command

CRLs have their own deployment section:

After: CRL Deployment section in CRL and OCSP

Each CRL binding controls its destination, format, optional parent bundle and reload command:

After: Attach CRL target dialog

## How it was verified

- full GitHub Actions **Tests & Quality Checks** workflow passed on commit `c72af3616`:
  - Frontend Tests
  - Backend Tests
  - Backend Tests (PostgreSQL)
  - Code Linting
  - Security Scanning
  - Docker Build Test
- targeted backend deployment, CRL and migration suite: 47 tests passed
- targeted frontend component suite after rebasing onto current `dev`: 4 files, 9 tests passed
- `git diff --check` passed
- installed and manually tested on UCM 2.231 in a Debian LXC
- verified revoke and revocation removal both regenerate and distribute the CRL
- verified PEM with and without parent CRLs and a single DER CRL
- verified CRL order and PEM object count
- verified two bindings receive the same update independently, including one failing target without blocking the other
- verified successful and failing reload commands, immediate delivery after editing, and queued retry after failure
- verified certificate binding paths can be cleared and binding-specific reload commands are retained
- verified bindings, options, paths and reload commands survive a complete UCM restart
- verified the CRL destination field starts empty and uses its example path only as a placeholder

## Compatibility

- migration 090 creates the CRL deployment tables
- migration 091 adds per-binding reload commands and preserves existing target commands on current bindings
- the old target reload column remains in the database for downgrade compatibility
- certificate and CRL deliveries continue to use the existing encrypted SSH credentials, host-key pinning, audit trail and retry model


NeySlim:
Thanks, this is a solid piece of work: the split between the reusable SSH connection and the per-binding paths and reload command is the right model, the durable per-target delivery is kept, the two migrations are dual-backend and idempotent, the backup manifest carries the new section with stable references, the CLA line is in, and the tests prove what they claim. I read the whole diff, ran the PR's backend tests without xdist, the frontend suite, the i18n sync check, and replayed migrations 090 and 091 on a PostgreSQL bench from an 089 schema, including the copy of a target's reload command into its binding. All green.

Eight points to fix before merge, all small and local, then it goes in.

**Backend**

1. **Deleting a certificate deletes CRL deliveries that share the binding number.** `services/cert/mixins/lifecycle.py:558-563` removes `DeployDelivery` rows by `binding_id` alone. Certificate and CRL bindings are two id sequences that both start at 1, so the collision is the norm, not the exception. Scenario: a pending CRL delivery for binding 2, delete the certificate behind certificate binding 2, the CRL delivery is gone and the CRL is not pushed until the next regeneration. Fix: filter on `binding_type == 'certificate'`, as `api/v2/deploy.py:131-136` already does.

2. **Deleting a CA that has a CRL binding returns 500.** `services/deletion_blockers.py` (`ca_deletion_blockers`, `purge_ca_dependents`) does not know `CRLDeployBinding`; the relation has no cascade, so SQLAlchemy tries `ca_id = NULL` on a NOT NULL column and `delete_ca` rolls back with "Failed to delete CA" on both backends. Fix: a 409 blocker naming the binding, like SCEP profiles and ACME domains, or purge the binding and its `crl` deliveries in `purge_ca_dependents`.

3. **A backup taken before this PR loses the reload command when restored.** `services/backup/restore/apply.py:130` only knows `include_root` as a legacy default. An older archive carries `deploy_targets.reload_command` and bindings without the column; after restore the binding has `reload_command = NULL`, and `services/deploy/service.py:244` only runs the binding's command, so pushes are "delivered" without any reload and without a message. Migration 091 does not replay on rows the restore inserts. Fix: when the archived binding row has no `reload_command`, take the one from the archived `deploy_targets` row it references; add the case to `tests/test_restore_legacy_defaults.py`.

4. **Changing a remotely executed command is no longer audited.** `update_binding` (`api/v2/deploy.py:298`) and `update_crl_binding` (`:472`) have no `AuditService.log_action`, while every other mutation in the file has one and the module docstring promises it. Before the PR the command lived on the target and its PATCH was audited. Fix: audit both PATCH routes with the changed fields.

**Frontend**

5. **CRL binding status badge.** `CRLDeploySection.jsx:147-149` shows the raw backend status (`delivered`, `pending`, `failed`) without `t()`, in the success colour for anything but `failed`, and shows "Enabled" when there is no delivery even if the binding is disabled. `CertDeploySection.jsx:161-163` has the right pattern (`deploy.status.*` keys and `STATUS_VARIANT`). Same fix as the certificate section.

6. **The delivery error is not shown.** The CRL section does not display `last_delivery.last_error` nor `delivered_at`; the certificate section does (`CertDeploySection.jsx:195-204`). When the automatic push after a CRL regeneration fails, the operator sees a red badge and no cause.

7. **`common.format` does not exist in any locale** (`CRLDeploySection.jsx:243`); only the English fallback renders. `export.format` exists, or add the key to the nine files.

8. **The `crlDeploy` block is English in seven locales.** Only `de.json` is translated; fr, es, it, ja, pt, uk and zh carry the English text word for word. The sync check is structural and passes; the content needs the seven translations (19 keys). `deploy.reloadCommandHint` in the same seven locales still carries the old wording ("one fixed command") while en and de have the new one.

**Nits, take or leave**

- `include_parent_crls` is read with `bool()` (`service.py:638`) while `include_root` goes through `json_boolean`: `"false"` becomes true.
- A binding with `include_parent_crls` on a CA whose parent has no CRL is accepted (201) and then fails on every delivery and every regeneration; validate the parents at create and PATCH, or skip parents without a CRL and say so.
- `crl.updated` reaches webhooks subscribed to `*` but is not in the event catalogue.
- `delete_crl_binding` lacks the try/except of its twin `delete_binding`; `models/deploy.py:193` lists three event types, there are six now.
- Help text: the Settings > Deployment help still describes the reload command on the target, and the CRL & OCSP help has nothing on the new section.
- The edit modal title is built by concatenation (`Edit: name`); an interpolated key like `deploy.attachTargetFor` reads right in every language.
- No CHANGELOG entry; please add one under Unreleased > Added (CRL deployment, per-binding reload commands, binding edit) and one under Changed (target settings reduced to the connection, migration 091 copies the command).

Once the eight points are in, I merge into dev for 2.233.

PR #363, opened 2026-09-21
feat(ad): one domain controller per row, with failover and health checks

stefanelul2000:
## Problem

The Active Directory Connector could not verify LDAPS against a domain with
more than one domain controller.

Its single Server field invites the domain name. A bare domain round-robins
across every DC, while TLS still verifies the certificate against that domain
name - and a DC certificate issued from the stock Domain Controller template
carries only its own FQDN in the SAN. So verification failed whichever DC
answered:

    socket ssl wrapping error: certificate ... 'subjectAltName':
    (('DNS', 'DC2.domain.com'),) doesn't match any name in ['domain.com']

Naming one DC worked; the domain never did.

## Change

The connector now takes several domain controllers, one per row.

- `_connect` walks them in order and returns the first that binds, so a single
  DC being down no longer stops certificate enrollment.
- Test Connection binds to **every** row and reports its result against that
  row, rather than stopping at the first DC that answers - a second DC that is
  broken would otherwise stay invisible until the first one failed.
- A scheduled probe binds to each DC on its own cadence, so an unreachable one
  is known before a client needs it instead of being rediscovered by an
  enrollment that waits out `_LDAP_CONNECT_TIMEOUT_SECONDS`. An unreachable DC
  is then skipped while it stays down.

Storage stays in the existing `server` column, newline-separated, so a config
saved with one hostname parses as a one-element list and keeps working. No
schema change for the DC list itself.

## Logging

Transitions only - a DC down for an hour does not rewrite the same warning
every cycle.

| Event | Level |
|---|---|
| A DC becomes unreachable | `WARNING`, naming it and why |
| The **last** reachable DC drops | `ERROR` - no naked-CSR subject can be derived and no Enroll ACL can be evaluated, so autoenrollment is failing for every client |
| Recovery | `WARNING` |

## Design notes for review

Three decisions worth pushing back on if you disagree:

1. **The probe is a full bind, not a TCP connect.** The failure this connector
   was built for - a DC certificate whose SAN names only its own host - passes
   a TCP check and fails the TLS handshake. A shallower probe would report a DC
   healthy that cannot serve a lookup.

2. **Health is advisory and expires.** `health.prioritise` may reorder and thin
   the candidate list, never empty it, and it is ignored once older than
   `_HEALTH_STALE_AFTER_SECONDS` (900s). If the probe itself breaks - a UCM-side
   DNS blip marking everything down - the connector still tries and the bind
   decides. A broken monitor must not become an outage.

3. **The interval is read from the row on each wake**, not from the value passed
   to `register_task`. Only the worker holding the scheduler's singleton lock
   runs the loop, so re-registering from whichever web worker handled the
   settings save would update a copy that never fires. Changing the interval now
   takes effect on the next wake with no restart. Floor is 60s, since the
   scheduler wakes once a minute and anything lower is a number it cannot honour.

Also fixed along the way: `_build_tls`'s temp CA bundle was unlinked the moment
the first connection returned. With failover there are several attempts, and
each later one needs the file still on disk. `cleanup()` now runs after the loop.

`health` is excluded from backups alongside `last_test_at` / `last_test_result`
- it is this installation's view of which DCs answered, meaningless on another.
`health_probe_interval` is config and is backed up.

## Verification

- **179 backend tests** pass, run in small batches on the rebased branch: 96 AD
  connector (42 of them new - splitting, failover order, exhaustion, CA-file
  lifetime, per-DC results, health precedence, staleness, the never-empty
  guarantee, `since` carry-forward, interval changes), 39 backup manifest/export,
  44 WSTEP/CSR subject.
- **Migration 093** applied both columns added, existing row preserved, interval defaulted to 120,
  idempotent on a second pass, no-op on a database without the table. Re-tested
  after the renumber.
- **Exercised live** on a dev instance: task registered, two unreachable DCs
  probed, `WARNING` per DC and `ERROR` on the last one, state stored as `down`
  with per-DC `since`, and the next scheduler tick returned in 0.3ms without
  re-probing (interval honoured, no repeat logging).
- **UI confirmed** against a real two-DC domain: both rows bound, per-row ticks
  rendered, summary row reads `dc1… +1 more`.
- i18n parity re-checked after the rebase: no keys lost from dev, exactly 10
  added (4710 → 4720).

## Not verified

- **PostgreSQL migration path** - no PG instance available; the SQLite path was
  tested directly.
- `test_restore_secret_storage.py` has 3 pre-existing failures, identical on
  clean `dev` and on the pre-rebase base. Unrelated to this branch - the test
  writes through `security.encryption.encrypt_text()` without requesting
  conftest's `encryption_enabled` fixture, so its secret is already cleartext
  before any restore.


stefanelul2000:
CLA is signed in the other PR. Please let me know if it's required here as well.

NeySlim (changes requested):
Thanks for this one. The failover and the per-DC test are the right shape, the migration is idempotent on both paths, the backup exclusion holds, and the single-hostname compatibility is real. I ran the full backend suite sequentially on the branch: green (the three `test_restore_secret_storage.py` failures you mention do not reproduce here). Three things to fix before merge, then a few smaller ones.

### Blocking

1. **`servers` that is not a string or a list of strings returns 500** (`backend/api/v2/ad_connector.py` `_parse_servers`, `backend/services/ad_connector/lookup.py` `split_servers`). `PUT /api/v2/ad-connector {"servers":[123]}` raises `AttributeError`; `{"servers":5}` or `{"servers":{"a":1}}` raise `TypeError` in `re.split`. Reject non-string entries and a `raw` that is neither `str` nor list, so the API answers 400.

2. **The staleness window never scales with the interval** (`health.py`: `_HEALTH_STALE_AFTER_SECONDS = 900` against `MAX_PROBE_INTERVAL_SECONDS = 86400`). With any interval above 15 minutes, `is_fresh` is false for most of the cycle, so `prioritise` returns the list untouched and the verdict never reaches `_connect`. Meanwhile the settings badge keeps showing Degraded, since it ignores `checked_at`. Either derive the window from the row (`max(900, 2 * interval)`) or cap the interval at 900 s, and make the badge respect the same window.

3. **`run_health_probe` does not speak the scheduler's contract** (`health.py` vs `scheduler_service.py` `_outcome_of`). It returns `{'skipped': ...}` and `{'error': ...}`; the scheduler expects `{'status': 'skipped'|'failed', 'reason': ...}` (see `services/backup/schedule.py`). Every 60 s wake is therefore logged "completed successfully" and counted as a successful run in the admin view, including on an installation with no connector and including when `probe()` raised.

### Should fix

- `split_servers` splits a string on commas and newlines, but not the elements of a list: `{"servers":["dc1.corp.local,dc2.corp.local"]}` is stored as one entry and read back as two. The inline test result is then keyed on the joined string while the probe keys on each host, and re-saving the same form sees `joined != config.server` and clears `health`. Apply the same split to each list element.
- `test_connection_saved` calls `health.record` even when the connector is disabled: a manual test on a disabled connector with unreachable DCs logs the ERROR "Kerberos autoenrollment will fail" and stores `state=down`. Gate `record` (or at least the transition log) on `config.enabled`.
- `SettingsPage.jsx` `handleAdConnectorTest` shows a success toast on a `partial` result; `AdConnectorForm.jsx` shows a warning for the same case. Same branch in both.
- The bind password is now `required` on first save (`AdConnectorForm.jsx`), which it was not on dev, and nothing in the API asks for it. Either drop it or say so in the CHANGELOG.
- No upper bound on the number of DCs beyond the 500-char join. N black-holed DCs block the scheduler's single thread for N × 10 s per cycle, delaying CRL, backup and discovery tasks. A cap of about 8 in `_parse_servers` is enough.
- `if not ok: db.session.rollback()` after `safe_commit` in `run_health_probe`: `safe_commit` already rolled back, and the function still returns the state as if it had been stored. Return `{'status': 'failed', ...}` there.
- `adConnector.server` is now an orphan i18n key in all nine locales.
- Migration 093 on PostgreSQL guards with `inspect(conn).get_columns`; 090 to 092 use `ADD COLUMN IF NOT EXISTS`. Idempotent either way, please align for consistency.

### CHANGELOG

Both entries run to five sentences; the project keeps entries to one or two (what was broken, what it does now). Please add the attribution `(#363, by @stefanelul2000)` as the neighbouring entries do, and restore the blank lines between `### Added`, `### Changed`, `---` and `## [2.232]` that the diff removes.

One note on the description: the file has 92 AD connector tests, not 96.

The CLA check will pass on its own once #364 (which carries your signature) is merged; nothing to do here.

stefanelul2000:
Thanks, all eleven are addressed, in `555dfca96` with the unrelated test guard kept separate in `7f9ca346a`. Per item:

## Blocking

**1. `servers` that is not a string or a list of strings returns 500.**
Fixed on both sides. `split_servers` (`backend/services/ad_connector/lookup.py`) now skips entries that are not strings and treats a `raw` that is neither `str` nor `list`/`tuple` as empty. It is called on the enrollment path and on a column that could have been hand-edited, so it has to degrade to "no DCs" rather than raise. The shapes it skips are then refused at the edge, in `_parse_servers` (`backend/api/v2/ad_connector.py`), which both `PUT /api/v2/ad-connector` and `POST /api/v2/ad-connector/test` go through: `{"servers":[123]}`, `{"servers":5}` and `{"servers":{"a":1}}` all answer 400. Without that second half a bad list would have stored silently with the bad entries dropped, which is worse than the 500.

**2. The staleness window never scales with the interval.**
Derived from the row, rather than capping the interval: a daily probe is a legitimate setting for a stable domain, and capping it at 900 s would have cost that installation 96 binds a day to keep a verdict it does not need that often. `_HEALTH_STALE_AFTER_SECONDS` is now `_HEALTH_STALE_FLOOR_SECONDS`, under `stale_after(interval) = max(900, 2 * interval)`. `is_fresh` and `prioritise` take the interval, and `_connect` passes `config.health_probe_interval`.

The badge reads the same window rather than its own copy of it: `to_dict` carries `health_stale_after` alongside `health`, and `AdConnectorSection.jsx` drops the badge once `health.checked_at` is older than it. So the badge cannot claim Degraded for a verdict `_connect` has already stopped acting on.

**3. `run_health_probe` does not speak the scheduler's contract.**
It now returns `{'status': 'ok' | 'skipped' | 'failed', 'reason': ...}`, matching `services/backup/schedule.py`. No connector, or disabled, or not due yet gives `skipped` with a reason, a `probe()` that raised gives `failed`, and anything else gives `ok` with the state counts. The 60 s wakes that do nothing are no longer logged and counted as successful runs.

## Should fix

**4. `split_servers` does not split the elements of a list.**
Every entry is split on commas and newlines now, list or string, the two paths differing only in how the entries are obtained. `["dc1.corp.local,dc2.corp.local"]` stores as two servers, so the inline test's key and the probe's key agree, and re-saving the same form no longer looks like a change and clears health.

**5. `test_connection_saved` calls `health.record` on a disabled connector.**
`record` is gated on `config.enabled`. The response still carries the full per-DC breakdown, and `last_test_at`/`last_test_result` are still written. Only the stored verdict, which drives the badge and the skip list, is withheld for a connector that serves no lookups. Covered by a test that asserts `config.health == {}` while `last_test_result` is set.

**6. Success toast on a partial result in `SettingsPage.jsx`.**
`handleAdConnectorTest` warns on `partial` now, same branch and same message as `AdConnectorForm.jsx`.

**7. Bind password required on first save.**
Kept, and the API now enforces it too, which is the half that was missing. An empty bind DN or password is an anonymous bind: ldap3 makes one whenever either half is missing, and this connector reads the computer and user objects that become the subject of a certificate UCM is about to issue. A directory that permits the anonymous read makes that identity attacker-influenced, and one that does not turns every lookup into a silent failure, so neither is a state worth being able to save. `PUT /api/v2/ad-connector` answers 400 when the save would leave the connector without either half, rolling back the partial update, and `POST /api/v2/ad-connector/test` refuses to test a bind it will not let you save. The form asked for both already; the point of the server-side check is that calling the route directly can no longer skip it.

A connector that already has a password is unaffected: the form never re-sends it, a blank field still means "unchanged", and the field is only `required` while none is stored. Clearing either half explicitly (`{"bind_password": ""}`, `{"bind_dn": ""}`) is refused. CHANGELOG entry added under Changed, as you asked.

**8. No upper bound on the number of DCs.**
`MAX_SERVERS = 8` in `_parse_servers`, refused with a 400 and a comment naming why the number matters (N black-holed DCs times the connect timeout, on the scheduler's single thread, ahead of CRL, backup and discovery).

**9. `db.session.rollback()` after `safe_commit`.**
Removed. That branch returns `{'status': 'failed', 'reason': ...}`, because the verdict in hand is not the one `_connect` will read, so reporting it as the run's result would claim a probe that did not land.

**10. Orphan `adConnector.server` key.**
Removed from all nine locales; `npm run check:i18n` is green.

**11. Migration 093 vs 090 to 092.**
Switched to `ADD COLUMN IF NOT EXISTS` on the PostgreSQL path. I kept the table-existence guard: migrations run before `create_all`, so on an installation that predates the connector the table is not there yet and its model builds it with these columns already present.

## CHANGELOG

Both entries are down to two sentences (what was broken, what it does now), carry `(#363, by @stefanelul2000)` like their neighbours, and the blank lines around `### Added`, `### Changed`, `---` and `## [2.232]` are back. A third entry under Changed covers the bind DN and bind password now required by item 7.

## Notes

- The per-row verdict inside the dialog is still shown regardless of age. It names its own timestamp ("Unreachable since ...") rather than asserting a current state, so I left it outside the freshness window. Happy to gate it too if you would rather the two agree exactly.
- On the count: that file collects 93 tests on the commit you reviewed, not 92 or 96. It is 99 now, plus 15 in `test_ad_connector.py`, so 114 across the two files.
- `pytest tests/test_ad_connector.py tests/test_ad_connector_lookup.py`: 114 passed. They cover the server-list shapes, the interval-scaled window, the scheduler contract including the failed-probe and not-due paths, the 400s for the bad shapes and the cap, the disabled-connector gate, and the bind credential rules above. eslint, the frontend build and `check:i18n` are clean.
- Full backend suite on this branch, SQLite, run sequentially, nothing else running: 7369 passed, 40 skipped, 0 failed. The fortieth skip is the IPv6 test in the next note, on a host that has no `::1`.
- One change in this branch is not about the connector, so it is a commit of its own (`7f9ca346a`) that you can drop or take separately: `tests/test_listen_address.py::test_dual_stack_listener_accepts_ipv4_and_ipv6` guarded itself with `socket.has_ipv6`, which only reports how the interpreter was built and stays true on a kernel with IPv6 switched off. On such a host the bind of `[::]` still succeeds, so the existing skip does not fire either, and the failure lands later as `EADDRNOTAVAIL` on the client's connect to `::1`, reading like a broken dual-stack listener. It now probes by binding `::1` itself, which is the address the client dials. It still runs wherever IPv6 works, CI included.
- Agreed on the CLA check and #364, nothing done here.

NeySlim (changes requested):
Second pass on `555dfca96` and `7f9ca346a`. All eleven points are addressed, and each was checked by running it: the 400s on malformed `servers`, the window derived from the interval on both `_connect` and the badge, the scheduler contract through `_outcome_of`, the cap, and migration 093 replayed on a real PostgreSQL. Full backend suite sequential on the branch: green. `7f9ca346a` is fine to take as is.

Item 7 introduces one regression, and the rule it adds stops halfway. That is the only blocking item left.

### Blocking

**A connector saved without a password can no longer be enabled or disabled.** The settings toggle sends `{"enabled": false}` alone, and `update_config` now answers 400 `bind_password is required` because the stored row has none. Such rows exist: the password was not required before this branch. Reproduced: row with `server`, `bind_dn`, `enabled=True`, no password; `PUT {"enabled": false}` gives 400 and `enabled` stays true.

Meanwhile `test-saved`, the scheduled probe and `_connect` still bind anonymously on that same row (`cfg.bind_password is None`, 200, health recorded), so the rule "an anonymous bind is not a connection this connector is allowed to make" only holds for `POST /test`.

Please make it one rule, applied everywhere:

- The PUT requires bind DN and password only when the resulting config is `enabled`. Disabling a legacy connector goes through, and step-by-step setup stays possible until the switch is turned on.
- `_connect`, `probe` and `test_connection_saved` refuse to bind without both, with one clear log line (for example `AD Connector: bind credentials missing, refusing anonymous bind`), instead of falling back to an anonymous bind.
- One upgrade sentence in the CHANGELOG entry: an enabled connector saved without a password stops answering lookups until credentials are entered in the settings dialog, and an API client must send both on the first save. AD refuses anonymous binds by default, so this should reach nobody in practice, but it has to be written down.

### Small things

- `probe()` says "Never raises" but `lookup._build_tls` sits outside its `try`. If it raises (the temp file cannot be created), every 60 s wake returns `failed` with two ERROR lines, whatever the interval. Move it inside, and drop the `logger.error` in `health.py` since the scheduler already logs a failed run.
- The cap test derives its input from `MAX_SERVERS` and does not check the message: with `MAX_SERVERS = 100` it still passes through the 500-character bound. Nine literal hosts and an assertion on `at most 8` would pin it.
- `AdConnectorSection.jsx` computes freshness with the browser clock. A client clock off by more than the window hides or keeps the badge wrongly. Emitting `health_fresh` from `to_dict` next to `health_stale_after` avoids that; not blocking.
- On item 3, one correction to the reply: the `skipped` wakes are still logged at INFO by the scheduler every minute and still increment `run_count`. That is the scheduler's own convention, nothing to change here, but the description should not say they are no longer logged.
- The comments above the credential check in the route and in the form run to ten lines each; the project keeps code comments to one to three lines stating the invariant. The reasoning belongs in the commit message.

stefanelul2000:
All in.

The PUT only asks for the credential when the resulting config is enabled, so `{"enabled": false}` on a passwordless row goes through and setup can be filled in over several saves.

`_connect`, `probe` and `test_connection` all refuse to bind without both halves, one line each: `AD Connector: bind credentials missing, refusing anonymous bind`. `_connect` raises rather than logs, since its callers already log that message, and the scheduled probe skips rather than warning every wake. One thing fell out of it: on a refused bind test-saved passed the empty server list to `health.record` and overwrote a real verdict, so it records only when DCs were actually tried.

Small ones: probe catches `_build_tls` and the `logger.error` is gone. It got its own try rather than the existing one, whose `finally: cleanup()` would NameError on the call that just raised. Cap test is nine literal hosts asserting on `at most 8`. `to_dict` emits `health_fresh` next to `health_stale_after` and the badge reads it.

CHANGELOG has the upgrade sentence, and the two comments are down to three lines and two.

On item 3 you're right, nothing in the repo said it. My reply overstated it.

Backend suite sequential: 7386 passed, 40 skipped. Frontend 100 files / 1080 tests, lint and i18n clean.

NeySlim (changes requested):
Third pass on `104c15856`. The credential rule is right now, and each half has a test that dies when I remove it: enabling a passwordless row is refused while disabling one goes through, setup can be filled in over several saves, and `_connect`, `probe`, `test_connection` and the saved test all refuse to bind anonymously. The saved test no longer overwrites a stored verdict with an empty one, which is a good catch of yours. Full backend suite sequential on your branch: 7388 passed, 38 skipped. No conflict with `dev`, which has moved on since you branched (a templates merge and an HSM fix, neither touching the connector).

One thing to fix, then this is ready.

### Blocking

**A `_build_tls` that raises is now recorded as a verdict instead of a failed run.** `probe` catches it and returns `[]`, and `run_health_probe` cannot tell "nothing was probed" from "every DC answered nothing": it calls `record(config, [])`, which stores `state: unknown` with an empty `servers` map and a fresh `checked_at`, commits, and returns `{'status': 'ok'}`. The real verdict and every per-DC `since` are lost, the dialog goes blank, the badge disappears, and the scheduler counts a green run that never reached a domain controller. On a host where the temp directory is read-only it repeats every interval.

Reproduced: enabled connector with a stored `degraded` verdict, `_build_tls` patched to raise `OSError`, result `{'status': 'ok', 'state': 'unknown'}` and the stored health replaced.

Before `555dfca96` the exception propagated and the verdict survived, so this is a regression of the fix rather than the pre-existing behaviour. Either let `probe` signal the failure (raise, or return `None`) and have `run_health_probe` answer `{'status': 'failed', 'reason': ...}` without calling `record`, or drop the inner `try` and keep the one in `run_health_probe`. On that last point: `tls, cleanup = _build_tls(config)` sits outside the `try/finally`, so the `NameError` you were guarding against cannot happen there.

### Small

- `test_connection` in `lookup.py` says "Never raises" but has the same shape: `_build_tls` is outside its `try`, so a failed temp file gives a 500 instead of `success: False`. Pre-existing, not from this branch, fine as a follow-up.
- The `Changed` CHANGELOG entry runs to four sentences; the project keeps them to one or two. The ldap3 mechanism belongs in the commit message, the rule and the upgrade sentence in the entry.
- That entry says the requirement applies "in the API as well as in the settings form" and that filling one in field by field still works. The form still has `required={!config?.bind_password}` regardless of `enabled`, so field-by-field setup is an API path only. Either reword it or tie the form's `required` to the enabled switch.

stefanelul2000:
I hope I got it right now :)

`probe` returns `None` when it could not run at all, `[]` still means there was nothing to probe, and `run_health_probe` answers `{'status': 'failed'}` on `None` without going near `record`. Took that option rather than dropping the inner try so "Never raises" stays true. Your reproduction is a test now: put the `return []` back and it fails with `assert 'ok' == 'failed'`.

You're right about the NameError, and I was guarding a shape nobody proposed. Moving the call inside the existing try would have left `cleanup` unbound in the `finally`; leaving it where it is, as you said, has no such problem. I should have checked which one you meant.

CHANGELOG entry is two sentences, the rule and the upgrade sentence. The ldap3 mechanism is in the commit message of the commit that took it out of the entry.

For the form I tied `required` to the switch rather than rewording, since the entry's claim is worth making true: `bind_dn` and `bind_password` now ask exactly what `update_config` asks. There was no AdConnectorForm test at all, so I added one in the shape of the other settings tests; reverting the form to the unconditional `required` kills two of its four cases.

PR #364, opened 2026-09-21
Feature/templates system custom

stefanelul2000:
## Problem

The templates list decided whether a template was a "Certificate" or a "CA" in
the browser, from a substring match on its name:

```js
if (name.includes('ca') || name.includes('authority') || t.is_ca || t.basic_constraints?.ca)
  return 'ca'
```

The three fallbacks are dead - `to_dict` returns none of `type`, `is_ca` or
`basic_constraints` - so every template fell through to the name match. And
`'ca'` catches those letters anywhere:

| Name | Shown as | Why |
|---|---|---|
| Web Server Certifi**ca**te | **CA** | `certificate` |
| Smart**ca**rd Logon | **CA** | `smartcard` |
| Client Certifi**ca**te | **CA** | `certificate` |
| Web Server (TLS/SSL) | Certificate | - |

Same kind of template, opposite labels, one word apart. Verified against a live
instance: not one template has `template_type = 'ca'`, not one carries
`keyCertSign`. All three "CA" badges were fiction.

The guess also drove the row icon, the badge colour, the mobile card badge, the
detail-pane badge and the detail-pane Type field, which spelled it out in full
as "Certificate Authority" - five elements agreeing with each other because they
read the same bad guess.

It could not mis-issue anything (issuance reads `extensions_template`), but on a
PKI tool, telling an operator that three of their templates create certificate
authorities is a serious thing to say wrongly.

## Context

`migrations/017_remove_ca_template.py` already deleted the built-in CA template:

> CA certificates should be created from the CAs page, not via templates in the
> Certificates page. This removes the misleading CA template.

That decision was made; this classifier is the orphan left behind.

## Change

The column now reads `is_system`, which the API already returns, and says
**System** or **Custom** - the difference between those templates that is real.
`getTemplateType` is deleted, along with all five places that echoed it.

- **System templates no longer offer Edit and Delete and then fail.** The server
  has always answered 403 to both (`templates.py` on update,
  `template_deletion_blockers` on delete). Those buttons are greyed with the
  reason on hover. Duplicate and Export stay available - the copy is created with
  `is_system: False`, which is the supported way to base a custom template on a
  built-in one, and it is what the disabled Edit tooltip points at.
- **A Show system toggle** hides the built-in templates. It remembers the choice
  in `localStorage`, and forces itself on - disabled, with a tooltip - while no
  custom template exists, so the list is never empty for want of a filter. The
  override is applied on top of the stored preference and deliberately **never
  written back**, so hiding system templates, then deleting your last custom
  template, does not destroy the choice.

## Shared component

`ResponsiveDataTable` gains two additive things, both opt-in:

- `disabled` / `disabledReason` on a row action - greyed, inert, reason on hover.
- A `toggle` filter type, for the show/hide case where a `multiSelect` with one
  option reads as a filter to unset rather than a switch. A toggle returns
  `false` from the "filters are active" check and the chip builder, so it does
  not light up the cleared-filters indicator or produce a chip whose Clear would
  fight the remembered preference.

No other caller passes either, so existing tables take the same paths as before.

## Also removed

`TemplatesPage` built a `rowActions` callback and never passed it to the table -
true on `dev` too, so long-standing dead code rather than a regression. Found by
a test asserting on buttons that turned out never to render. Deleted (29 lines);
every identifier it used still has live references from the detail panel.

One consequence to be explicit about: **`disabled` row actions are now used by no
page.** The support is correct and has its own test, but on Templates the guard
lives entirely in the detail panel - backed by the server's 403s, which predate
all of this.

## Verification

13 frontend tests under the project's vitest 4.1.11:

- `TemplatesPage.systemCustom.test.jsx` (8) - the sharpest is a same-name pair,
  `Identical Certificate A/B`, differing only by `is_system`. Nothing about the
  text can decide that, so it pins the API contract: if `is_system` stops
  arriving, every row silently reads Custom and every system template gets its
  buttons back. Plus the disabled/enabled action matrix and all four toggle
  behaviours including the never-write-back invariant.
- `dataTableDisabledActions.test.jsx` (5) - a disabled action does not fire, is
  really `disabled` rather than only styled, and the flag is per row.

`check-i18n-sync.js` clean across all nine locales after the rebase (4722 keys,
no keys lost from dev). UI confirmed manually in both toggle states.


NeySlim (changes requested):
Thanks, the diagnosis is right and the `is_system` column is the correct fix. `to_dict` returns `is_system` on every route feeding this page, the server does answer 403 on PUT and DELETE for system templates, and duplicate creates the copy with `is_system: False`. Translations are real in all nine locales. A few things before merge.

### The red CI job

`floatingWindowRbac.test.jsx` reads the source text of `TemplatesPage.jsx` and looks for `const rowActions =`, which you removed. You are right that the callback was dead code on dev. That test guarded a menu nobody ever saw; the real guard is the detail panel, which your change keeps. Please replace the failing case with a behavioural one, for instance with the mocks of your `TemplatesPage.systemCustom.test.jsx` and a `usePermission` mock driven by a variable:

```jsx
it('offers template edit and duplicate only on write:templates', async () => {
  perms.write = false
  getAll.mockResolvedValue({ data: [CUSTOM] })
  render(<MemoryRouter><TemplatesPage /></MemoryRouter>)
  fireEvent.click(await screen.findByText(CUSTOM.name))
  await screen.findByRole('button', { name: /common.export/i })
  expect(screen.queryByRole('button', { name: /common.edit/i })).toBeNull()
  expect(screen.queryByRole('button', { name: /common.copy/i })).toBeNull()
})
```

Removing `canWrite('templates') &&` from the panel makes it fail, which is what the old test never proved.

### Blocking

1. **Filter presets no longer apply** (`TemplatesPage.jsx` `handleApplyFilterPreset`): it still reads `filters.type` while the filter key became `source`. A preset saved after this change does nothing; a preset saved before it (`{type:['ca']}`) empties the table.

2. **The persisted filter keeps its old values** (`usePersistedState('ucm-filter-templates-type', ...)`): a user who had "Certificate" selected before upgrading gets an empty list after, because `certificate` matches neither `system` nor `custom`. Either use a new storage key or drop unknown values on load.

3. **"Clear filters" turns the toggle off and writes `""` to localStorage** (`ResponsiveDataTable.jsx`, the empty-state clear button and `handleClearAllChips`): both loops call `f.onChange('')` on every filter, including the toggle, so `setShowSystemPersisted('')` hides the system templates and stores an empty string (read back as "on" on the next load). Skip `type === 'toggle'` in both, and in `getCurrentFilterValues`, which currently records `showSystem: true` into every saved preset.

### Should fix

- The two CHANGELOG entries sit under `## [2.232]`, which shipped yesterday, instead of `## [Unreleased]`. They also run to four sentences each (the project keeps entries to one or two), and the second says the buttons are greyed "in the row and in the detail pane alike", while the row has no actions on this page, as your description itself notes.
- `helpContent` (dead, `helpPageKey` is what renders) still describes Certificate and CA badges that no longer exist.
- Switching the toggle off leaves the detail panel open on a system template that is no longer listed.
- Two of the eight new tests (`leaves a custom template fully actionable`, `does not overwrite the stored preference`) also pass against dev; the other six do fail there, so the contract is pinned, but those two prove nothing on their own.
- The example names in the description ("Web Server Certificate", "Client Certificate") are not the seeded names. The bug is real on three seeds (Email Certificate, Client Authentication, Smartcard Logon), which is worth stating as is.

stefanelul2000:
## Update: review feedback addressed

### The red CI job

`floatingWindowRbac.test.jsx` no longer reads `TemplatesPage.jsx` as text. The replacement is behavioural, in `TemplatesPage.systemCustom.test.jsx`, with a `perms` variable driving the `usePermission` mock, essentially as you wrote it, plus a mirror case for `delete:templates`:

```js
it('offers template edit and duplicate only on write:templates', async () => {
  perms.write = false
  ...
  expect(screen.queryByRole('button', { name: /common.edit/i })).toBeNull()
  expect(screen.queryByRole('button', { name: /common.copy/i })).toBeNull()
})
```

Removing `canWrite('templates') &&` from the panel now fails. The comment left in `floatingWindowRbac.test.jsx` records why the old shape could not fail for the reason that mattered.

### Blocking

**Filter presets.** `handleApplyFilterPreset` reads `filters.source`. A preset saved before the rename carries `{type:['ca']}`, falls to the `else`, and clears the filter rather than emptying the table. Both paths have a test, the second applied on top of a narrower preset so a handler that ignored it would be caught.

**The persisted filter.** New storage key, `ucm-filter-templates-source`. The state is `filterSource` now rather than `filterType`, so the variable and the key cannot drift apart again the way they just did.

**Clear filters.** `type === 'toggle'` is skipped in `handleClearAllChips` and in `getCurrentFilterValues`. The empty-state button no longer carries its own copy of the loop, it calls `handleClearAllChips`, so there is one clearing rule instead of two that have to agree. New `dataTableToggleFilter.test.jsx` covers both surfaces, the preset save, and that a toggle is not counted as an active filter.

### Should fix

Both changelog entries moved to `[Unreleased]`, cut to two sentences each, with the real seeded names and no claim about row actions.

`helpContent` is gone, along with the `HelpCard` import it was the only user of.

Switching the toggle off now closes the detail panel if the open template was a system one.

The two tests that also passed against dev now pin the new contract. The custom-template one asserts the pane prints the Source description and that the buttons carry no refusal title; the stored-preference one re-renders with a custom template added and asserts the remembered `false` still hides the system rows.

You were right about the example names, and the correction is worth more than a rename. The three seeded templates the bug actually hit are **Email Certificate (S/MIME)**, **Client Authentication** and **Smartcard Logon**, each containing those two letters in certifi**ca**te, authenti**ca**tion and smart**ca**rd. The test fixtures use those names now.

### Two things your review led to

**The rendered help said the same vanished thing.** You flagged the dead `helpContent`; the two files that do render described End-Entity and CA templates as well. `data/helpContent.js` now has a Source section reading System and Custom, plus entries for the real Type values and the Show system toggle. `data/helpGuides.js` replaces its "Template Types" section, and its step 3 of Creating a Template no longer says "select the template type (End-Entity or CA)", which no dialog has ever offered. Seven locale keys orphaned by the two deletions are removed from all nine files; `check:i18n` reports them in sync at 4715 keys.

**The `rowActions` `disabled` support is removed.** Your point that the row has no actions on this page made me check who passes it: nobody. Every `rowActions` call site in `frontend/src` was scanned for a `disabled` key and the only hits were in its own test file. So the plumbing and that test file are gone, and `ResponsiveDataTable`'s row-action code is now byte-identical to dev. The only thing this branch adds to that component is the toggle filter type and the clear and save handling it needs. The detail panel keeps its own disabled Edit and Delete, which is where a user actually meets them.

### Verification

Full suite: 102 files, 1097 tests, all passing. `eslint src/` and `npm run build` clean, `check:i18n` in sync.

NeySlim (changes requested):
Second pass on `b9ded3f42` to `02821355c`. Everything from the first review is fixed and was checked by mutation: the behavioural RBAC test dies when the `canWrite` guard is removed, the toggle handling dies when either exclusion is removed, the preset handler dies on the old key, the two tests that passed against dev now fail there, and the row-action code of `ResponsiveDataTable` is back to dev byte for byte. Full frontend suite: 102 files, 1097 tests, green. `check:i18n` in sync.

Two things to change, one to leave alone.

### Please leave the help files to me

`helpContent.js` and `helpGuides.js` are the English source of the Quick Help, and the eight other languages live in `data/help/<lang>/templates.js`, merged by section index at render time. Your rewrite of the English side is right, but with the other eight untouched the panel is now wrong for every non-English user: the first section still reads "Template Types / End-Entity / CA", the new "Type" item shifts the feature list so "Show system" appears in English and Import/Export twice, and the guide keeps "CA Templates" and step 3 "End-Entity or CA". You had no way to see this, the sync checker for those files is not in the repository.

Please revert `e05bf1ffc` and `02821355c` (the help rewrite and the seven keys it orphaned; the keys can go with the resync). I will do the nine-language resync as a follow-up right after the merge, so the English text you wrote is not lost, it will land with its translations.

### Blocking

**`applies a preset saved on the Source filter` does not test the handler.** Its `waitFor(() => expect(queryByText('Smartcard Logon')).toBeNull())` is satisfied on the first tick by the table's transient "Loading..." render, before the filtered rows come back. The test passes against dev and with `handleApplyFilterPreset` reading `filters.type`. Wait for a positive state first (`await screen.findByText(CUSTOM.name)`, then assert the system row is absent), or stabilise `showError` in the `useNotification` mock so `useCRUDPage` stops refetching on every render. The same `waitFor(toBeNull)` shape is used in two other cases; they happen to be caught by their other assertions, but the pattern is worth avoiding.

### Small

- The two CHANGELOG entries lack the attribution `(#364, by @stefanelul2000)` that the neighbouring entries carry.

stefanelul2000:
Reverted both rather than dropping them (`590f5564f`), so the help is exactly as it was and you get a fast-forward. 

You were right about the preset test, and the real problem was a level above the `waitFor`. The `useNotification` mock handed back a fresh object on every call, but the real context memoises it, so `loadData` got a new identity each render and refetched in a loop. The table was flipping back to `Loading...` constantly, which is what the absence check was passing on. Stabilised the mock and reshaped the assertions to wait for the custom row and read the system row out of the same render, scoped to the table (`f7e1f07ac`). It fails now if the handler reads `filters.type`. Did the same to the other two, and had a look at `dataTableToggleFilter` and `floatingWindowRbac` while I was in there, neither has the pattern.

Attribution added to both entries (`1fadabef3`).

Frontend suite and `check:i18n` both green.

Issue #286, opened 2026-08-15
FR:  Certificate identifiers


arthur-2sheds-jackson:
EE certificates are currently identified in the UI by their CN/Common Name.  The CAB forum gives a "not recommended" for this field in such certs, and LE currently issues certificates under their "tlsserver" & "shortlived" profiles without this field set.

I'd like to request a string identifier ("Name"? "Identifier"?) for certificates, non-unique & mutable, maybe defaulting to the first domain listed in the SAN. 

This would both allow certificates without a value in their CN field to be identified in a manner similar to the current CN method, and to distinguish certificates that are outwardly identical.

Currently, a cert using an RSA key has nothing to distinguish it's entry in the "Certificates" list from anothe for the same domain but using an ECDSA key except for the "Key Type" field, which isn't shown in certificate selectors.  Being able to identify a cert by "Web Server" or "Servicename (RSA)"/"Servicename (ECDSA)" in a selector would be easier than guessing between multiple certs identified as "host.example.com"

Edit: The impetus for this was the list of ACME-issued certs, but applies to both certs issued by the system (in it's capacity as a CA) and those requested by the system (from ACME/an outside CA).

NeySlim:
Implemented in dev — ece276360 (will ship with 2.211):

- Certificates get a mutable, non-unique display name, independent from the CN — **Rename** action in the certificates list (API: `PATCH /api/v2/certificates/<id>` with `{"descr": "..."}`).
- It defaults to the CN, or the first SAN DNS name for CN-less certs — both for certs issued by UCM as a CA (including its ACME server) and those obtained through the ACME client (which already used the primary domain).
- Lists and certificate selectors show this name first, and selectors now append the key type (`RSA 2048` / `EC secp256r1`) — so the "same domain, RSA vs ECDSA" case is distinguishable even without renaming anything.

Note the certificate list already fell back CN → first SAN → description for display, so LE `shortlived`/`tlsserver` certs were not fully anonymous — the mutable label and the key type in selectors are the new parts.

arthur-2sheds-jackson:
Thank you!

NeySlim:
Shipped in **v2.211** — mutable display names with SAN-based defaults for CN-less certs, and key type shown in certificate selectors. Thanks for the well-argued request.

PR #365, opened 2026-09-22
fix(certificates): show Common Name and add Description column

B0F1B0:
## What this changes

The Certificates table displayed the certificate description under **Common Name** because row normalization preferred `descr` over `cn` and `common_name`.

For example, a certificate issued with Common Name `this-is-the-common-name.lan` and description `THIS-IS-THE-DESCRIPTION` showed `THIS-IS-THE-DESCRIPTION` in the Common Name column.

This change:
- Removes the description-first override so the Common Name column uses the existing CN fields.
- Adds a separate **Description** column that can be shown or hidden through **Show Columns**, with the visibility preference remembered.
- Supports Description sorting using the existing backend `descr` sort field.
- Displays an em dash for empty descriptions and reuses the existing translations.

The API's existing fallback behavior for certificates without a CN is unchanged. No backend or database changes are required.

## How it was verified

- 7/7 tests passed in `CertificatesPage.filterAfterAction.test.jsx`, including separate CN/description values, Description sorting and hiding Description while keeping the CN visible.
- Confirmed that the new regression test fails with the original description-first normalization.
- Tests passed again after rebasing onto the current `dev` branch.
- Production frontend build passed.
- Targeted ESLint checks reported 0 errors (existing warnings only); all nine locales are in sync.
- Installed the frontend patch on a UCM 2.233-rc2 LXC. Service and health checks passed, and the served frontend bundle matched the installed file.
- The corrected UI was manually verified on the LXC.

### Creating Example Certificate with Common Name and Description

Screenshot 1 - Certificates - Issue Certificate

## UI before

Screenshot 2 - Certificates - Common Name

## UI after

Screenshot 4 - Certificates - Correct Common Name and new Column Description

- [x] I confirm that commits with an unlinked author email are my own work, or that I am authorised by their copyright holder to submit and license them under the UCM CLA v1.1.


NeySlim (changes requested):
Thanks, the observation is right: a description typed at issuance, or the `SCEP: <subject>` and ACME domain the enrolment paths store, ends up in the Common Name column. The test is good too, it fails when the description-first line is put back.

But that line is not a bug. It came with the mutable display name from #286 (`ece276360`): the **Rename** action on this page edits `descr` so that the name an operator chooses shows in the main column, for certificates with no CN in particular. With this change, renaming a certificate no longer changes anything in the main column, and the pickers (`CertificatePickerModal`, `CertificateInput`, `CompareModal`) still show `descr || common_name`, so the list and the pickers stop naming the same certificate the same way.

Two features share one field, and the fix has to keep both. What I would like instead:

- The main column shows the CN on the first line, and `descr` underneath as a secondary line when it is set and differs from the CN. A certificate with no CN keeps `descr` on the first line, as today.
- Rename keeps working as it does, its effect stays visible in the list, and the pickers stay consistent with it.
- No separate Description column: the subtitle covers it, and the sort on `descr` can stay wired to the main column's secondary line if you want to keep it.

Mobile render in `useCertificateColumns.jsx` has the same fallback chain and should follow.

Small: the CHANGELOG entry lacks the attribution `(#365, by @B0F1B0)` the neighbouring entries carry.

NeySlim:
Thanks! The changes asked in the review were small, so rather than another round we made them ourselves: merged into `dev` with a follow-up commit on top, which keeps the Rename action visible in the list:

- The first column shows the Common Name on the first line and the description under it when it is set and differs. A certificate without a CN is still named by its first SAN, then its description, as the API already does.
- No separate Description column: the subtitle covers it, and the CSV export writes `CN (description)` in that column.
- Your test is replaced by three that cover the subtitle on desktop and mobile, the description equal to the CN, and the export.

It will ship in 2.233.

